PR #29, opened 2020-01-08
add: new cmake option for standalone static library

ayounes-synaptics:
BOOST_JSON_STANDALONE: new cmake option to enable compilation of
boost::json as a static standalone dependency without any
dependency, it also also compiles the examples.

warning: C++-17 is mandatory in this mode

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,13 @@ else()
     target_compile_definitions(boost_json PUBLIC BOOST_JSON_STATIC_LINK=1)
 endif()
 
-if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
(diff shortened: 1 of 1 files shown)


vinniefalco:
Wonderful. This is the first time I'm actually happy about seeing a cmake `option` !!

ayounes-synaptics:
The failed test seems to be an environment issue:

gpg: requesting key BA9EF27F from hkp server keyserver.ubuntu.com
Error: retrieving gpg key timed out.
gpg: keyserver timed out
gpg: keyserver receive failed: keyserver error
The command "sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test"" failed and exited with 1 during .

vinniefalco:
> The failed test seems to be an environment issue: 

Yeah, usually when I restart the build errors like that go away. I'll do it now.

vinniefalco:
voila

ayounes-synaptics:
> voila

Merci :-)

PR #33, opened 2020-01-10
fixes parse of -1e-308

madmongo1:
error was in `finish()` in `exp3` state


include/boost/json/detail/impl/number.ipp
```diff
@@ -92,9 +92,16 @@ pow10(int exp) noexcept
 
         1e+300, 1e+301, 1e+302, 1e+303, 1e+304, 1e+305, 1e+306, 1e+307, 1e+308 };
 
-    exp += 308;
-    BOOST_ASSERT(exp >= 0 && exp < 618);
-    return tab[exp];
+    if (exp < -308 || exp > 308)
+    {
+        return std::pow(10.0, exp);
```
vinniefalco: idk about this

include/boost/json/detail/impl/number.ipp
```diff
@@ -290,11 +290,12 @@ loop:
                         n_.u = m;
                         goto enter_mantd;
                     }
-                    ++p;
+                    ++p; // consume digit
```
vinniefalco: Another way to spell "consume digit" is `++p`

include/boost/json/detail/impl/number.ipp
```diff
@@ -141,6 +141,8 @@ maybe_init(char ch) noexcept
     else
     {
         dig_ = 1;
+        // a leading non-zero digit is always significant
+        sig_ = 1;
```
vinniefalco: extraneous comment

include/boost/json/detail/impl/number.ipp
```diff
@@ -351,7 +353,8 @@ loop:
                     // VFALCO Check dig_ for overflow
                     // Could use an implementation-defined limit
                     // which is lower than USHRT_MAX
-                    ++dig_;
+                    ++sig_; // increase number of significant digits in mantissa
+                    ++dig_; // increase total number of digits in mantissa
```
vinniefalco: extraneous comments

test/_detail_number.cpp
```diff
@@ -421,136 +452,139 @@ class number_test
...
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000" // 500 zeroes
+        for (int i = 0 ; i < 2 ; ++i)
```
vinniefalco: FYI This is valid C++
```
for(auto b : {true, false})
```

test/_detail_number.cpp
```diff
@@ -395,20 +418,28 @@ class number_test
...
-            std::memcpy(&ugot, &got, sizeof(got));
-            log << 
+            log << "not close : " <<
+                f.name() << "\n"
+                "string: " << s << "\n"
+                "need  : " << diagnose(need) << "\n"
+                "got   : " << diagnose(got) <<
+                std::endl;
+        }
+        else if (pedantic and !BOOST_TEST(same))
```
vinniefalco: `&&`

vinniefalco:
The first line of a commit message should be short, certainly not with a number having 500 zeroes...

There's no need to quote the number in the commit message, since it is already visible in the test.

CMakeLists.txt
```diff
@@ -87,4 +87,11 @@ else()
         PUBLIC
             Boost::system
     )
+    option(BUILD_TESTING "Build the tests" ON)
```
vinniefalco: Commit message should read
"Add cmake option to build tests and examples"

include/boost/json/detail/impl/number.ipp
```diff
@@ -299,6 +310,9 @@ loop:
                 }
                 if(*p == 'e' || *p == 'E')
                 {
+                    // if decimal point has not yet been seen, assume it's here
+                    // since 100e0 is equivalent to 100.0e0
+                    if (pos_ < 0) pos_ = dig_;
```
vinniefalco: The statement has to be on its own line
```
if (pos_ < 0)
    pos_ = dig_;
```
for two reasons: to be able to set a breakpoint, and to distinguish the lines in coverage reports.

include/boost/json/detail/impl/number.ipp
madmongo1: I'll get to it. I wanted to prevent the assert from happening

test/_detail_number.cpp
madmongo1: ah, nice.

test/_detail_number.cpp
madmongo1: oops. old habits...

vinniefalco:
This should be squashed down into just one commit

CMakeLists.txt
vinniefalco: Outdated review comment, disregard

include/boost/json/detail/impl/number.ipp
vinniefalco: Do we still need this check?

include/boost/json/detail/impl/number.ipp
```diff
@@ -215,7 +223,7 @@ loop:
             BOOST_ASSERT(dig_ == 0);
             ++p;
             pos_ = 0;
-            st_ = state::zeroes;
+            st_ = state::mant_frac_1_digit;
```
vinniefalco: Suggested: `state::mantf`

include/boost/json/detail/impl/number.ipp
```diff
@@ -230,7 +238,35 @@ loop:
         goto finish;
     }
 
-    //-----------------------------------
+    // 1*digit
```
vinniefalco: Good comment

include/boost/json/detail/impl/number.ipp
```diff
@@ -230,7 +238,35 @@ loop:
...
+                st_ = state::zeroes;
+                ++p;
+                ++dig_;
+            }
+            else
+            {
+                if(neg_)
+                    st_ = state::mantn;
+                else
+                    st_ = state::mant;
```
vinniefalco:
```
else if(neg_)
{
    st_ = state::mantn;
}
else
{
    st_ = state::mant;
```

include/boost/json/detail/impl/number.ipp
```diff
@@ -230,7 +238,35 @@ loop:
...
+            }
+            goto loop;
+        }
+        else
+        {
+            ec = error::expected_fraction;
+            goto finish;
+        }
```
vinniefalco: No need for `else` since we used `goto` in the `if`

include/boost/json/detail/impl/number.ipp
vinniefalco: The comment is extraneous, `++p` already means "consume digit"

include/boost/json/detail/impl/number.ipp
```diff
@@ -282,11 +318,12 @@ loop:
...
                         goto enter_mantd;
                     }
-                    ++p;
+                    ++p; // consume digit
                     // VFALCO Check dig_ for overflow
                     // Could use an implementation-defined limit
                     // which is lower than USHRT_MAX
-                    ++dig_;
+                    ++dig_; // increase digit count
+                    ++sig_; // increase significant digit count
```
vinniefalco: Extraneous comments

include/boost/json/detail/impl/number.ipp
```diff
@@ -295,10 +332,15 @@ loop:
...
                     n_.kind = kind::double_;
-                    continue;
+                    st_ = state::mant_frac_1_digit;
+                    n_.u = m;
+                    goto loop;
                 }
                 if(*p == 'e' || *p == 'E')
                 {
+                    // if decimal point has not yet been seen, assume it's here
+                    // since 100e0 is equivalent to 100.0e0
```
vinniefalco: Comment could be simplified to `// 'E' counts as '.'

include/boost/json/detail/impl/number.ipp
```diff
@@ -348,10 +391,18 @@ loop:
...
+                    n_.u = m;
+                    st_ = state::mant_frac_1_digit;
+                    goto loop;
                 }
                 if(*p == 'e' || *p == 'E')
                 {
+                    // if we see an exponent but have not yet seen a decimal point
+                    // then the decimal point must be assumed to be at the position of
+                    // the [eE]
+                    // e.g. 100e0 is equivalent to 100.0
```
vinniefalco: LOL what is this wall of text? No one wants to read that, and no one will read that... consider `// 'E' counts as '.'` instead

include/boost/json/detail/impl/number.ipp
```diff
@@ -396,6 +447,10 @@ loop:
             }
             if(*p == 'e' || *p == 'E')
             {
+                // if we have not encountered a decimal point then
+                // we must assume it's at the [eE]
+                // because 100e0 is equivalent to 100.0e0
+                if (pos_ < 0) pos_ = dig_;
```
vinniefalco: The condition and statements need to be on separate lines for coverage and breakpoints

include/boost/json/detail/impl/number.ipp
```diff
@@ -636,16 +694,32 @@ finish(
...
-        exp_ += off_;
-        if(eneg_)
+    {
+        if (eneg_)
             exp_ = -exp_;
+
+        if (pos_ == 0)
+        {
+            // this indicates that the decimal point preceeds any significant digits
+            // which means the mantissa has a magnitude < 1.0
```
vinniefalco: Consider just `// mantissa < 1`

include/boost/json/detail/number.hpp
```diff
@@ -39,6 +39,7 @@ class number_parser
     enum class state
     {
         init,   init0,  init1,
+        mant_frac_1_digit,      // mantissa: after decimal there must be a digit
```
vinniefalco: `mantf` would be more in line with the style, and there's no need for the comment here since there is one in the case label (the `1*digit`)

test/_detail_number.cpp
```diff
@@ -16,6 +16,8 @@
 
 #include "test_suite.hpp"
 
+#define BOOST_JSON_DETAIL_NUMBER_DEMAND_PERFECT_DOUBLE_CONVERSION 0
```
vinniefalco: Not sterile, and this macro is only visible in tests so you can just use `#define ACCURATE_CONVERSION` or something like that

test/_detail_number.cpp
```diff
@@ -255,16 +356,22 @@ class number_test
...
         check_bad("0.0+");
         check_bad("0.0e+");
         check_bad("0.0e-");
         check_bad("0.0e0-");
         check_bad("0.0e");
+        check_bad("0.e1"); //must be at least 1 digit after mantissa decimal
         check_bad("-e");
         check_bad("-x");
+        check_bad("2.e+3"); //must be at least 1 digit after mantissa decimal
+        check_bad("-2.e+3"); //must be at least 1 digit after mantissa decimal
```
vinniefalco: These comments are all extraneous, `check_bad` subsumes them

test/_detail_number.cpp
```diff
@@ -544,19 +657,8 @@ class number_test
...
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000"
-            "00000000000000000000000000000000000000000000000000" // 500 zeroes
-            "1e600", f_boost{} );
+        check_bad("-2.e+3"); //must be at least 1 digit after mantissa decimal
```
vinniefalco: extraneous comment

include/boost/json/detail/impl/number.ipp
madmongo1: we could just check for "more than 3 digits" and then rely on the finalize to detect the over/underflow. I think i'd prefer that...

oh wait. What if the input is 0.[ten thousand zeroes]1e10000 ?
there's no reason we shouldn't parse this correctly.

It would be trivial to compute the exponent limits on entry into this state

test/_detail_number.cpp
madmongo1: no longer needed. we're not parsing accurately atm

test/_detail_number.cpp
madmongo1: gone now

include/boost/json/detail/impl/number.ipp
madmongo1: may I at least comment the purpose of the state?

include/boost/json/detail/impl/number.ipp
madmongo1: I'd want to read it :)

Resolved with `// 'E' implies '.' if not already encountered`

include/boost/json/detail/impl/number.ipp
madmongo1: will change to `// |mantissa| < 1.0` as this succinctly contains 2 more data:

1. we are considering the magnitude of the mantissa
2. the magnitude of the mantissa is a real number

include/boost/json/detail/impl/number.ipp
madmongo1: In my view the check is superfluous. it's more expensive to perform the check than to fail late.

vinniefalco: In general I think it is good to fail as soon as we know we have an invalid input. So if we get 309 as the exponent, we can fail right there. I'm not happy about having a potential call to `std::pow`, especially when we know that the exponent is out of range.

include/boost/json/detail/impl/number.ipp
```diff
@@ -230,7 +237,34 @@ loop:
         goto finish;
     }
 
-    //-----------------------------------
+    // 1*digit
+    case state::mantf:
```
vinniefalco: if you really want to make up for the comment removed from the header, you could write
```
// 1*digit
// A digit after the decimal point is required.
case state::mantf
```
Although it isn't strictly needed since the BNF already explains it.

include/boost/json/detail/impl/number.ipp
```diff
@@ -295,10 +330,15 @@ loop:
...
                     pos_ = dig_;
                     n_.kind = kind::double_;
-                    continue;
+                    st_ = state::mantf;
+                    n_.u = m;
+                    goto loop;
                 }
                 if(*p == 'e' || *p == 'E')
                 {
+                    // 'E' implies '.' if not already encountered
```
vinniefalco: "if not already encountered" is superfluous, it is already spelled `if (pos_ < 0)`. This comment could be more succinct if written as `// treat 'E' as '.'`

include/boost/json/detail/impl/number.ipp
```diff
@@ -348,10 +389,15 @@ loop:
...
                     pos_ = dig_;
                     n_.kind = kind::double_;
-                    continue;
+                    n_.u = m;
+                    st_ = state::mantf;
+                    goto loop;
                 }
                 if(*p == 'e' || *p == 'E')
                 {
+                    // 'E' implies '.' if not already encountered
```
vinniefalco: `// treat 'E' as '.'`

include/boost/json/detail/impl/number.ipp
```diff
@@ -396,6 +442,9 @@ loop:
             }
             if(*p == 'e' || *p == 'E')
             {
+                // 'E' implies '.' if not already encountered
```
vinniefalco: `// treat 'E' as '.'`

include/boost/json/detail/impl/number.ipp
```diff
@@ -636,16 +688,31 @@ finish(
...
     case state::exp3:
-        exp_ += off_;
-        if(eneg_)
+    {
+        if (eneg_)
             exp_ = -exp_;
+
+        if (pos_ == 0)
+        {
+            // |mantissa| < 1.0
```
vinniefalco: I like this comment, and it can be even better written as `abs(mantissa) < 0`

include/boost/json/detail/impl/number.ipp
```diff
@@ -636,16 +688,31 @@ finish(
...
+            auto start = dig_ - sig_;
+            auto exponent_adjust = -start - 1;
+            exp_ += exponent_adjust;
+        }
+        else
+        {
+            // mantissa >= 1.0
```
vinniefalco: is this `abs(mantissa) >= 1` ?

include/boost/json/detail/number.hpp
```diff
@@ -48,7 +49,6 @@ class number_parser
     number n_;
     short exp_; // exponent string as integer
     short dig_; // digits in mantissa
-    short off_; // mantissa's contribution to exponent
     short pos_; // position of decimal point
     short sig_; // significant digits in mantissa
```
vinniefalco: good comments

test/_detail_number.cpp
```diff
@@ -36,12 +38,111 @@ operator==(
...
+            throw;
+        }
+    }
+
+    static void emit_scientific(std::ostream& os, double d)
+    {
+        auto oldflags = os.flags();
+        try
+        {
+            constexpr auto digits = std::numeric_limits<double>::max_digits10;
```
vinniefalco: what is `constexpr` buying us here and why even have `digits` at all? i.e. why not
```
std::setprecision(std::numeric_limits<double>::max_digits10)
```
?

When we name a thing we elevate its importance, we should not elevate things needlessly.

test/basic_parser.cpp
```diff
@@ -481,7 +481,10 @@ class basic_parser_test
                 s.data(), s.size(),
                 ec);
             if(! BOOST_TEST(! ec))
+            {
+                ::test_suite::log_type() << "    failed to parse: " << s;
```
vinniefalco: Not a fan of this why didn't we keep it using the `log` data member?

madmongo1: hmm it failed to compile for me for some reason. Will recheck

include/boost/json/detail/impl/number.ipp
madmongo1: 0.0000000001e309 is actually 1e289. stod will parse this correctly.

test/basic_parser.cpp
vinniefalco: the lambda needs to capture `this`, and functions which use the log or call functions (or lambdas) that use the log cannot be `static`

vinniefalco:
I think we should merge the `mant` and `mantn` states as you proposed today

test/_detail_number.cpp
madmongo1: The reason I do this:
Spelling out concepts one per line is to me the same as your preference for formatting code in terms of one piece of entropy per line.
Naming a temporary variable like this is a zero-cost abstraction. `constexpr` is a signal that this is a 'number concept' as opposed to an actual value object.

Nevertheless, your code, your rules. Will change.

include/boost/json/detail/impl/number.ipp
```diff
@@ -636,16 +688,31 @@ finish(
...
+    {
+        if (eneg_)
             exp_ = -exp_;
+
+        if (pos_ == 0)
+        {
+            // abs(mantissa) < 1
+            auto start = dig_ - sig_;
+            auto exponent_adjust = -start - 1;
+            exp_ += exponent_adjust;
```
vinniefalco: despite all the naming fanfare it produces warnings:
```
2>C:\Users\vinnie\src\boost\libs\json\include\boost/json/detail/impl/number.ipp(705,36): warning C4244: '+=': conversion from 'int' to 'short', possible loss of data
2>C:\Users\vinnie\src\boost\libs\json\include\boost/json/detail/impl/number.ipp(711,36): warning C4244: '+=': conversion from 'int' to 'short', possible loss of data
```

vinniefalco: When you take away the unnecessary local variables and simplify the formulas, the interesting relationship between the values becomes visible:
```
        if (pos_ == 0)
        {
            // abs(mantissa) < 1
            exp_ = static_cast<short>(
                exp_ + sig_ - dig_ - 1);
        }
        else
        {
            // abs(mantissa) >= 1
            exp_ += static_cast<short>(pos_ - sig_);
        }
```

ofenloch: I don't know if my problem is related to your discussion. I have a simple Catch2 test for boost.json 1.75.0 that fails because float/double values are not written/read properly:

```cpp
TEST_CASE("boost::json")
{
    // There seems to be an issue with number formatting.

    // create a boost::json::value
    boost::json::value jv_original = {
        {"pi", 3.14159265359},
        {"e", 2.71828182846}};

    // serialize the boost::json::value to a string
    std::string s{boost::json::serialize(jv_original)};
    // and write the string to a file
    std::string file_name = std::tmpnam(nullptr);
    file_name += ".json";
    std::ofstream ofs(file_name);
    ofs << s;
    ofs.close();

    // now read the file into a string
    std::ifstream ifs(file_name);
    std::string content((std::istreambuf_iterator<char>(ifs)),
                        (std::istreambuf_iterator<char>()));
    ifs.close();
    // and create another boost::json::value from that string
    boost::system::error_code parseErr;
    boost::json::value jv_unserialized = boost::json::parse(content, parseErr);

    // compare both boost::json::value objects
    CHECK(jv_original == jv_unserialized);

} // TEST_CASE("boost::json")
```

The error message from Catch2 is
```bash
[ctest]   CHECK( jv_original == jv_unserialized )
[ctest] with expansion:
[ctest]   {"pi":3.14159265359E0,"e":2.71828182846E0}
[ctest]   ==
[ctest]   {"pi":3.1415926535899996E0,"e":2.71828182846E0}
```

grisumbras: Please test against recent Boost version. I can't reprodcue your issue locally on 1.81.0

ofenloch: Thanks for your answer, @grisumbras.

You're right. I built and installed Boost 1.81.0 (still running on Ubuntu 20.04) and the failure is gone.

However, I use Boost.JSON 1.81.0 in one of my classes and still have the same problem with this value in a JSON file

```json
"latitude": 79.824228,
```

Putting this value in the code above results in success. But my compare(lhs, rhs) method says:

```bash
[ctest] values of "latitude" differ:
[ctest]   lhs 7.9824228E1
[ctest]   rhs 7.98242E1
```

I'll investigate further ...

PR #37, opened 2020-02-08
Implement new string interface

sdkrystian:
This implements the new `boost::json::string` interface, along with checked replacement and insertion. `replace` and `insert` for `InputIterator` still need to be implemented as checked.

### DIFF
--- a/include/boost/json/detail/impl/string_impl.ipp
+++ b/include/boost/json/detail/impl/string_impl.ipp
@@ -1,5 +1,6 @@
 //
 // Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+// Copyright (c) 2020 Krystian Stasiowski (sdkrystian@gmail.com)
 //
 // Distributed under the Boost Software License, Version 1.0. (See accompanying
 // file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
@@ -13,10 +14,22 @@
 #include <boost/json/error.hpp>
 #include <boost/json/detail/string_impl.hpp>
 #include <cstring>
+#include <functional>
 
 namespace boost {
 namespace json {
 namespace detail {
+
+inline
+bool
+ptr_in_range(
+    const char* first,
+    const char* last,
+    const char* ptr) noexcept
+{
+    return std::less<const char*>()(ptr, last) &&
+        std::greater_equal<const char*>()(ptr, first);
+}
 
 string_impl::
 string_impl() noexcept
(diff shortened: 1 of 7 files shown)


include/boost/json/detail/impl/string_impl.ipp
```diff
@@ -12,6 +13,7 @@
 
 #include <boost/json/error.hpp>
 #include <boost/json/detail/string_impl.hpp>
+#include <boost/json/detail/impl/string_impl.hpp>
```
vinniefalco: I don't think you should be including this here, it is already included in `<boost/json/detail/string_impl.hpp>`

include/boost/json/impl/string.hpp
```diff
@@ -77,12 +79,48 @@ insert(
...
-        impl_.insert(off, tmp.size(), sp_),
+        impl_.insert_unchecked(pos, tmp.size(), sp_),
         tmp.data(),
         tmp.size());
-    return impl_.data() + off;
+    return *this;
+}
+
+// KRYSTIAN TODO: this can be done without copies when
+// reallocation is not needed.
```
vinniefalco: Fix the comment to explain that we should use a separate algorithm if iterator category is Forward or better

include/boost/json/impl/string.hpp
```diff
@@ -77,12 +79,48 @@ insert(
...
+
+    // We use the default storage because
+    // the allocation is immediately freed.
+    storage_ptr dsp;
+    detail::string_impl tmp(
+        first2, last2, dsp,
+        iter_cat<InputIt>{});
+    cleanup c{tmp, dsp};
+    traits_type::copy(
+        impl_.replace_unchecked(first - begin(), last - first, tmp.size(), sp_),
```
vinniefalco: long line..

test/string.cpp
```diff
@@ -1570,55 +1570,55 @@ class string_test
...
-        // insert(size_type, string const&, size_type, size_type)
-        {
-            fail_loop([&](storage_ptr const& sp)
-            {
-                string s(t.v1, sp);
-                s.insert(1, string(t.v2), 1, 3);
-                BOOST_TEST(s == std::string(
-                    t.v1).insert(1, t.s2, 1, 3));
-            });
+        //// insert(size_type, string const&, size_type, size_type)
```
vinniefalco: need a comment explaining why this is disabled, put your name in caps `// SDK` or something like that so the reader knows it is editorial.

test/string.cpp
```diff
@@ -2290,6 +2188,129 @@ class string_test
...
+                string s2(t.v2, sp);
+                BOOST_TEST(s2.replace(s2.begin(), s2.begin() + 4, t.v2.substr(4, 2)) == s1.replace(0, 4, t.v2.data() + 4, 2));
+            });
+
+            // outside, grow
+            fail_loop([&](storage_ptr const& sp)
+            {
+                std::string s1(t.v2.data(), t.v2.size());
+                string s2(t.v2, sp);
+                BOOST_TEST(s2.replace(s2.begin(), s2.begin() + 1, t.v2.substr(0)) == s1.replace(0, 1, t.v2.data(), t.v2.size()));
```
vinniefalco: This is hard to read

vinniefalco: Reformatted:
```
BOOST_TEST(
    s2.replace(
        s2.begin(),
        s2.begin() + 1,
        t.v2.substr(0)) ==
    s1.replace(
        0, 1,
        t.v2.data(),
        t.v2.size()));
```
Now that I can read it... we are using the iterator overload of `json::string` but the index overload of `std::string`? Why?

vinniefalco:
Looks great, minor style nits, but the tests need some work.

vinniefalco:
45 uncovered lines here:
https://codecov.io/gh/vinniefalco/json/src/834b1e29f463fc391652eb22ffbaacf13540ea74/include/boost/json/detail/impl/string_impl.ipp

test/string.cpp
sdkrystian: Reformatted in new commit.

> Now that I can read it... we are using the iterator overload of `json::string` but the index overload of `std::string`? Why?

They perform the same operation, it was just easier to copy the index based tests.

test/string.cpp
sdkrystian: Added

include/boost/json/impl/string.hpp
sdkrystian: Moved to multiple lines.

include/boost/json/impl/string.hpp
sdkrystian: Done

include/boost/json/detail/impl/string_impl.ipp
sdkrystian: Fixed; `ptr_in_range` was turned into an `inline` non-template function and moved into `detail/string_impl.hpp`

PR #41, opened 2020-02-11
allow CML to find boost with flag

madmongo1:
This change allows a standalone Boost.Json to optionally respect the normal `find_package(Boost)` mechanism for finding an installed instance of Boost.

**Existing workflow is not affected in any way.**

setting `-DBOOST_JSON_FIND_BOOST=ON` will require the CML to find Boost, while respecting `BOOST_ROOT` as normal.


vinniefalco:
Merged, thanks

PR #42, opened 2020-02-14
Improve coverage for string

sdkrystian:


include/boost/json/detail/impl/string_impl.ipp
```diff
@@ -252,7 +252,7 @@ replace(
     {
         const bool inside = detail::ptr_in_range(curr_data, curr_data + curr_size, s);
         // there is nothing to replace; return
-        if (inside && s - curr_data == pos && n1 == n2)
+        if (inside && std::size_t(s - curr_data) == pos && n1 == n2)
```
vinniefalco: Can we please use `static_cast` so it sticks out?

sdkrystian: Will address

vinniefalco:
Merged, thanks

PR #47, opened 2020-02-26
Add documentation Travis target

vinniefalco:


vinniefalco:
This didn't work so well. Or at all.

PR #49, opened 2020-03-07
Fix warning: unused variable ‘init’ [-Wunused-variable]

mloskot:
Pedantic avoidance of

```
../../libs/json/test/value_ref.cpp:177:29: error: expected primary-expression before ‘init’
  177 |             (void)init_list init =
      |                             ^~~~
```


vinniefalco:
Thanks!

PR #52, opened 2020-03-07
Add minimal CONTRIBUTING.md

mloskot:
Make it clear `git clone --recurse-submodules` is required, otherwise not all components of the project can be built (namely, benchmarks).


vinniefalco:
Thanks!

PR #54, opened 2020-03-08
Fix move construction of value from object

sdkrystian:


test/value.cpp
```diff
@@ -499,6 +499,13 @@ class value_test
                 BOOST_TEST(jv.is_object());
                 BOOST_TEST(*jv.storage() == *sp);
             }
+            {
+                // verify moved from
+                object obj{{"a", 1}, {"a", 1}};
```
vinniefalco: I would not use duplicate keys

test/value.cpp
```diff
@@ -499,6 +499,13 @@ class value_test
...
+                BOOST_TEST(obj.capacity());
```
vinniefalco: I would not depend on conversion to bool. In other words, spell this out: `obj.capacity() > 0`

test/value.cpp
```diff
@@ -499,6 +499,13 @@ class value_test
...
+                value jv(std::move(obj), {});
+                BOOST_TEST(!obj.capacity());
```
vinniefalco: I would spell this out: `obj.capacity() == 0`

sdkrystian: Copy, will apply.

PR #55, opened 2020-03-08
Add moving of strings from value_ref, fix value constructor

sdkrystian:


sdkrystian:
Tests are failing as these changes are applied separately from #54, will pass once it is applied.

vinniefalco:
If one commit requires the other then they should be part of the same pull request. One commit per bugfix doesn't mean you have to put each one into its own pull request.

include/boost/json/impl/value_ref.ipp
```diff
@@ -67,8 +68,17 @@ string_view
...
+    BOOST_ASSERT(
+        what_ == what::str ||
+        what_ == what::strfunc);
+    // this function is only used to create objects,
+    // and the string is used as a key. since the key
+    // type is string_view, we no longer care about 
+    // move semantics.
+    if (what_ == what::strfunc)
+        return *static_cast<const string*>(f_.p);
+    else
```
vinniefalco: I would lose the `else`, it reads better

test/value_ref.cpp
```diff
@@ -307,6 +307,39 @@ class value_ref_test
...
+            BOOST_TEST(a.data() == nullptr);
+        }
+        {
+            object a{{"a", 1}, {"b", 2}};
+            BOOST_TEST(a.capacity() > 0);
+            // move sets the table pointer to 
+            // nullptr, and capacity returns 0
+            // when the table pointer is null
+            array b{std::move(a), object()};
+            BOOST_TEST(a.capacity() == 0);
```
vinniefalco: I would remove these comments because they don't provide any additional information

include/boost/json/impl/value_ref.ipp
```diff
@@ -46,8 +46,9 @@ is_key_value_pair() const noexcept
         return false;
     auto const& e =
         *arg_.init_list_.begin();
-    if( e.what_ != what::str)
-        return false;
+    if(e.what_ != what::str &&
+        e.what_ != what::strfunc)
+            return false;
     return true;
```
vinniefalco: Yep this looks much better without the else, as it is absolutely clear that the function ends on line 52

test/value_ref.cpp
sdkrystian: Done

vinniefalco:
Merged, thanks!

PR #56, opened 2020-03-12
Update CMake configuration to clarify supported build scenarios

mloskot:
Attempt to further clarify what build scenarios are supported by CMakeLists.txt

------

/cc @qis 
Sorry to bother you, but I am quite puzzled by this `out-of-tree` case. When this can happen exactly?

```cmake
else()
    #
    # Out-of-tree, non-standalone build
    #
    find_package(Boost COMPONENTS system)
    target_link_libraries(boost_json
        PUBLIC
            Boost::system
    )
    option(BUILD_TESTING "Build the tests" ON)
    if (BUILD_TESTING)
        add_subdirectory(bench)
        add_subdirectory(example)
        add_subdirectory(test)
    endif()

endif()
```

If no standalone and no in-Boost is requested, then `elseif(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR` case happens.

I must be missing something, otherwise this could become

```cmake
else()
    message(FATAL_ERROR "Unsupported build scenario")
endif()
```


qis:
@mloskot I'd assume it's for when you want to build the JSON library from trunk while using a stable Boost distribution for dependencies. STANDALONE also means no boost deps, which are sometimes wanted or even necessary.

mloskot:
@qis Makes sense, thanks. I will test that mode and update the PR

vinniefalco:
Is there a possibility of changing one or more of the Travis (or Azure) matrix items to exercise these additional use cases?

mloskot:
> while using a stable Boost distribution for dependencies.

@qis Hmm, does this imply use of `cmake -DBOOST_ROOT=/where/boost/stable/install` ?

I'm still trying to distinguish it from

- `BOOST_JSON_FIND_BOOST`
- `{boost-root}/libs/json` which uses stage build of Boost

mloskot:
@vinniefalco 
> Is there a possibility of changing one or more of the Travis (or Azure) matrix items to exercise these additional use cases?

I'm adding such builds for GitHub Actions 
- https://github.com/mloskot/boost-json/actions
- https://github.com/mloskot/boost-json/tree/ml/add-github-actions

I find workflows, each configured in its own `.yml` file, a much better structure than a single behemoth YAML full of conditions and mixture of bash scripts.

qis:
@mloskot Yes, if boost is installed using the system package manager, vcpkg or conan, then there is no `stage` directory. They can probably be combined, not sure.

Perhaps it would make sense to use `find_package(… CONFIG)` here? But sometimes the generated configs don't work because they're not exported using CMake and rely on a specific b2 `layout` setting.

I'd leave it alone or remove it until somebody comes along who needs it and opens an issue. Hopefully, Boost.JSON will be officially accepted into Boost soon - then it won't be needed at all.

mloskot:
This is now finished for the three of supported build scenarios, with corresponding GitHub Actions to test them.

There are still two build scenarios that needs to be clarified and tested:
1. CMake build as part of Boost superproject (see `TODO` comments).
2. CMake build as out-of-tree, non-standalone build
    - this one is unclear, seems equivalent to `BOOST_JSON_FIND_BOOST` mode

mloskot:
@vinniefalco This can be closed as you've already merged the commits manually.

PR #57, opened 2020-03-13
Improve string exposition

sdkrystian:


doc/qbk/02_2_strings.qbk
```diff
@@ -12,22 +13,58 @@
 [section:strings Using Strings]
 
 Modifiable sequences of characters are represented using instances of type
-[link json.ref.boost__json__string `string`].
-These strings have an interface almost identical to `std::string`, and
-operate the same way, with these variations:
+__string__.
+These strings have an modified `string_view` based version of the `std::string` interface, and
```
vinniefalco: "an" -> "a"

vinniefalco: Perhaps `string_view` should be a link to cppreference

doc/qbk/02_2_strings.qbk
```diff
@@ -12,22 +13,58 @@
...
+operate the same way, except that:
 
-* __string__ is not a class template.
+* __string__ is not a class template,
+
+* __string__ only uses `char` as its character type,
+
+* redundant overloads for string operations have been replaced with a `string_view` based interface,
+
+* access to characters in the range `[size(), capacity()]` is permitted, and
```
vinniefalco: This should be a half-open interval `[ ... )`

doc/qbk/02_2_strings.qbk
```diff
@@ -12,22 +13,58 @@
...
 
 * __storage__ is used instead of __Allocator__.
 
-* New characters are 
+With the `string_view` based interface, operations requiring an input string are implemented as a single overload with
+a parameter of type `string_view`, and can accept most string-like objects in addition to other `string_view`s.
```
vinniefalco: You're using the word string and string_view way too many times it is like "string string and string string view string the string view the viewed string view stringed view" - blah

doc/qbk/02_2_strings.qbk
```diff
@@ -12,22 +13,58 @@
...
-memory allocations using the default storage, or a specified storage:
+[snippet_strings_4]
 
-[snippet_strings_1]
+More formally, `std::string` member function overloads that accept any of the
+following parameter combinations as an input string:
+
+* a `std::string` parameter, or
+
+* a `std::string` parameter and two `size_type` parameters that specify a substring within that string, or
```
vinniefalco: "string" three times in a sentence

doc/qbk/02_2_strings.qbk
```diff
@@ -12,22 +13,58 @@
...
+
+are replaced with with an overload accepting a `string_view` parameter.
+
+This design removes several redundant overloads from the interface. For example, the 11 overloads of `std::string::insert` are reduced
+to just 3 in __string__, while still providing identical functionality. In addition to these changes, overloads taking a `std::initializer_list<char>` parameter have been removed. Such overloads have little use,
+as they serve as little more than a wrappers for arrays, with an inefficient syntax:
+
+[snippet_strings_3]
+
+With the removal of overloads that specify parameters for a substring, a member function `subview` that returns a `string_view` referring to
```
vinniefalco: 60 cols max please

doc/qbk/02_2_strings.qbk
```diff
@@ -12,22 +13,58 @@
...
+a substring within the __string__ is provided to facilitate cheap substring operations:
```
vinniefalco: string...within...string...to..substring the string view thing string viewed string string TOO MANY STRINGS!! ARGH

doc/qbk/02_2_strings.qbk
```diff
@@ -12,22 +13,58 @@
...
+
+[snippet_strings_2]
+
+A __string__ may be constructed using the default storage without incurrening any
```
vinniefalco: "incurrening"

test/snippets.cpp
```diff
@@ -29,6 +29,73 @@ usingStrings()
...
+    }
+
+    {
+        //[snippet_strings_4
+
+        json::string str = "Boost.JSON";
+        json::string_view sv = str;
+
+        // all of these call compare(string_view)
+        str.compare(sv);
```
vinniefalco: make these `assert`s (not `BOOST_ASSERT`), e.g.:
```
assert( str.compare(sv) < 0 );
```

test/snippets.cpp
```diff
@@ -29,6 +29,73 @@ usingStrings()
...
+
+        str.compare(str);
+
+        str.compare("Boost");
+
+        struct string_viewish
+        {
+            operator string_view()
+            {
+                return "Sorta stringish";
```
vinniefalco: This wording needs to be sterile

test/snippets.cpp
```diff
@@ -29,6 +29,73 @@ usingStrings()
...
+        // all of these call compare(string_view)
+        str.compare(sv);
+
+        str.compare(sv.substr(0, 5));
+
+        str.compare(str);
+
+        str.compare("Boost");
+
+        struct string_viewish
```
vinniefalco: This wording needs to be sterile

test/snippets.cpp
```diff
@@ -29,6 +29,73 @@ usingStrings()
...
+        {
+            operator string_view()
+            {
+                return "Sorta stringish";
+            }
+        };
+
+        // even for this
+        str.compare(string_viewish());
```
vinniefalco: Do we really need to show this in the examples? All it shows is that C++ conversion operator works. It doesn't add any information about the library. And it will probably confuse some folks.

sdkrystian: Alright, remove that

test/snippets.cpp
sdkrystian: For the other examples specifying the post-operation state of the string I added asserts. No asserts are needed here though.

doc/qbk/02_2_strings.qbk
sdkrystian: Fixed.

doc/qbk/02_2_strings.qbk
sdkrystian: You're right, it's a little excess. I trimmed that line down.

doc/qbk/02_2_strings.qbk
sdkrystian: That's the formal spec of what actually happens, so it has to be redundant.

doc/qbk/02_2_strings.qbk
sdkrystian: Fixed.

mloskot: (off-topic) I'm new here, what is the workflow?
- @vinniefalco requests the change
- @sdkrystian makes the change
- who hits "Resolve conversation"? @sdkrystian after fixing or @vinniefalco after verifying it's been fixed?

doc/qbk/02_2_strings.qbk
mloskot: Here is my quick try for the whole section

```
More formally, member function overloads for `std::string` accept either of the following parameter sets:

* object of `std::string` type

* object of `std::string` type and two `size_type` parameters that specify position of the first character and length of the substring in the object

* object of type convertible to __string_view__

* object of type convertible to __string_view__ and two `size_type` parameters that specify position of the first character and length of the substring in the object

* `const_pointer` to null-terminated character string
```

doc/qbk/02_2_strings.qbk
vinniefalco: I don't use any of that I just leave individual comments.

mloskot: All are individual comments but if thread is not resolved, not closed, it's dangling post PR update and confusing reviews, what's been addressed.

vinniefalco:
Merged, thanks!

doc/qbk/02_2_strings.qbk
```diff
@@ -12,22 +13,66 @@
...
-memory allocations using the default storage, or a specified storage:
+[snippet_strings_4]
 
-[snippet_strings_1]
+More formally, `std::string` member function overloads that accept any of the
+following parameter combinations as an input string:
+
+* a `std::string` parameter, or
+
+* a `std::string` parameter and two `size_type` parameters that specify a substring, or
```
mloskot: > two `size_type` parameters that specify a substring

Two numbers can specify substring or range of string in at least two different ways.

vinniefalco: The docs are a work in progress, I imagine that we will continue to keep improving them over many pull requests. @sdkrystian

mloskot: @vinniefalco Understood. I just picked and followed the nitpicking approach.

sdkrystian: @mloskot  There's a careful balance between being pedantic and not boring the user. I could write these in standardese, but that would bore the user. Besides, this specification doesn't need to be specific as it outlines the overloads that are removed, so exact behavior is not needed.

mloskot: @sdkrystian Yes, the balance is important but AFAIU that part is "formal spec" where such balance does not quite apply (as it doesn't in API reference either). Anyway, I'm being anal and I expect a [function spec to be boringly precise](https://en.cppreference.com/w/cpp/string/basic_string/substr). Anyway, my question has been answered. Thx.

vinniefalco: The place for boring precision is in the javadoc. The exposition should be concise, elegant, and inspiring.

PR #58, opened 2020-03-14
Fix CMake configuration for tests

mloskot:
A couple of things are broken that basically make `ctest` is a no-op, always:

```
Test project /home/runner/work/boost-json/boost-json/build
No tests were found!!!
```

Mind you, it is a success (`0` return code)!

This PR fixes all of them.


vinniefalco:
Merged, thanks!

PR #60, opened 2020-03-17
Updates

vinniefalco:


.github/workflows/cmake-find-boost-install.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: CMake
+        run: |
+          cmake -B "$GITHUB_WORKSPACE/build" -DCMAKE_VERBOSE_MAKEFILE=ON -DBOOST_JSON_FIND_BOOST=ON
+      
+      - name: Build
+        run: |
+          cmake --build "$GITHUB_WORKSPACE/build" --config Release
+
+      - name: Test
+        run: |
```
Flamefire: just because I love saving horizontal space: You can use `working-directory:` instead of the `cd` and make all `run:` simple one-liners (without the pipe)

.github/workflows/cmake-with-boost-stage.yml
```diff
@@ -0,0 +1,83 @@
...
+          git clone -b $BOOST_BRANCH --depth 1 https://github.com/boostorg/boost.git $BOOST_ROOT
+          cd $BOOST_ROOT
+          git submodule update -q --init libs/headers
+          git submodule update -q --init tools/boost_install
+          git submodule update -q --init tools/boostdep
+          git submodule update -q --init tools/build
+
+      - uses: actions/checkout@v2
+        with:
+          path: ${{ steps.setenv.outputs.boost_root }}/libs/${{ steps.setenv.outputs.boost_self }}
```
Flamefire: You don't need outputs AND env: `${{env.BOOST_ROOT}}` works

CMakeLists.txt
```diff
@@ -14,7 +14,11 @@ if(BOOST_SUPERPROJECT_VERSION)
     set(BOOST_JSON_VERSION ${BOOST_SUPERPROJECT_VERSION})
 endif()
 
-project(boost_json VERSION ${BOOST_JSON_VERSION} LANGUAGES CXX)
+if (BOOST_SUPERPROJECT_VERSION)
```
Flamefire: Isn't this pointless given https://github.com/CPPAlliance/json/pull/60/files#diff-af3b638bc2a3e6c650974192a53c7291R13-R15

.github/workflows/cmake-with-boost-stage.yml
mloskot: Good tip, thanks

CMakeLists.txt
mloskot: Sorry, no idea what this is referring to, lines 13-15 are

```
## Introduction

This library provides containers and algorithms which implement JSON
```

.github/workflows/cmake-find-boost-install.yml
mloskot: Good tip, thanks.

CMakeLists.txt
Flamefire: Those lines are in the CML, Github doesn't show it by default as it is not part of the changeset AND jumps to the wrong file -.-:

```
if(BOOST_SUPERPROJECT_VERSION)
    set(BOOST_JSON_VERSION ${BOOST_SUPERPROJECT_VERSION})
endif()
```

mloskot: > Isn't this pointless given

🤷‍♂ ask @vinniefalco 

The `BOOST_SUPERPROJECT_VERSION` handling was there before and I did not aim to remove it. I only patched it a bit.

Flamefire: Yeah but the original version already took care of BOOST_SUPERPROJECT_VERSION and https://github.com/CPPAlliance/json/pull/60/commits/bbfb09b26c8acb105c02328a9222af2af384d94f added superflous BOOST_SUPERPROJECT_VERSION handling. So I'd just revert that commit. The original is more DRY

.github/workflows/cmake-with-boost-stage.yml
mloskot: Done #62

.github/workflows/cmake-find-boost-install.yml
mloskot: Done #62

CMakeLists.txt
mloskot: Oh, good catch!

Not sure what I was thinking about or it was left after some numerous experiments and CMakeLists.txt edits. Done in #62

PR #61, opened 2020-03-17
GA: Add checkout of submodules

mloskot:
Change missing from #56

------

Should fix this GA failure https://github.com/CPPAlliance/json/runs/515012932

Based on https://github.com/actions/checkout#Checkout-submodules


vinniefalco:
Merged, thanks

PR #64, opened 2020-03-27
Fix value_ref construction for const rvalues

sdkrystian:


test/value_ref.cpp
```diff
@@ -92,6 +92,13 @@ class value_ref_test
             string const s{};
             (value_ref)s;
         }
+
+        // test that value_ref can be constructed
+        // from a const rvalue
```
vinniefalco: This comment can be terse:
```
// construct from const rvalue
```

The rest is implied by context.

sdkrystian: Alright, good to go.

sdkrystian:
Fixed by https://github.com/CPPAlliance/json/commit/fdee3ceb71a04c3b5f1b548a9dda9ae8ebe04b4d

PR #76, opened 2020-04-22
Enable cmake install:

madmongo1:
When building outside of boost, boost_json now supports the install option of cmake.
This patch produces a full, correct install which obeys CMAKE_INSTALL_PREFIX and
creates the correct config.cmake files in order to construct the target Boost::json
This config will automatically find Boost::system using the standard cmake find_package
command


madmongo1:
I have updated the generated config script so that standalone boost_json does not cause the Boost package to be found or linked against

CMakeLists.txt
```diff
@@ -123,6 +124,7 @@ else()
     find_package(Boost COMPONENTS system)
     target_link_libraries(boost_json
         PUBLIC
+            Boost::container
             Boost::system
```
vinniefalco, line 128: Why do we link to system here but not on line 96

CMakeLists.txt
```diff
@@ -45,12 +51,15 @@ endif()
...
 add_library(Boost::json ALIAS boost_json)
-
 target_compile_features(boost_json PUBLIC cxx_constexpr)
 
 # TODO: For Boost superproject, this may need to be INTERFACE setting
-target_include_directories(boost_json PUBLIC include)
+# Installed targets have different include requirements
+target_include_directories(boost_json PUBLIC
+        "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+        "$<INSTALL_INTERFACE:include>")
```
pdimov, line 62: It's more idiomatic to use `include(GNUInstallDirs)` and then `${CMAKE_INSTALL_INCLUDEDIR}` here.

CMakeLists.txt
```diff
@@ -110,15 +124,18 @@ else()
     find_package(Boost COMPONENTS system)
     target_link_libraries(boost_json
         PUBLIC
+            Boost::container
```
pdimov, line 127: If this is going to link to `container`, the `find_package` probably also needs to be adjusted.

test/cmake_install_test/main.cpp
```diff
@@ -0,0 +1,13 @@
...
+// Distributed under the Boost Software License, Version 1.0.
+// See accompanying file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt
+
+#include <boost/json.hpp>
+#include <iostream>
+
+int main()
+{
+    boost::json::error_code err = boost::json::error::expected_colon;
+    std::cout << err.message() << std::endl;
```
pdimov, line 12: This doesn't exercise any functionality from the compiled library, so may hide link errors.

test/cmake_install_test/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+find_package(boost_json REQUIRED)
+
+add_executable(main main.cpp)
+target_link_libraries(main Boost::json)
+
+enable_testing()
+add_test(NAME main COMMAND main)
+
+add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -C $<CONFIG>)
```
vinniefalco, line 18: something is going on with your editor where the last line has no newline

CMakeLists.txt
madmongo1: These are not libraries that find_package would find. I assume they are created by the build super-project (I found the file like this)

CMakeLists.txt
madmongo1: it's linked on line 100

test/cmake_install_test/CMakeLists.txt
madmongo1: I think my editor removes trailing newlines by default. I assumed this was normal.

test/cmake_install_test/main.cpp
madmongo1: isn't the compiled version just an artefact of trying to reduce travis build times? Will anyone actually use it?

vinniefalco: No, the point of the compiled version is to reduce everyone's build time, by removing as much material as possible from the headers.

PR #81, opened 2020-04-27
Refactor monotonic_resource

### LINKED ISSUE #78: Add initial resource, clear function & maximum byte cap to monotonic_resource

sdkrystian:
Resolves #78

### DIFF
--- a/include/boost/json/detail/monotonic_resource.hpp
+++ b/include/boost/json/detail/monotonic_resource.hpp
@@ -1,0 +1,91 @@
+//
+// Copyright (c) 2020 Krystian Stasiowski (sdkrystian@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/vinniefalco/json
+//
+
+#ifndef BOOST_JSON_DETAIL_MONOTONIC_RESOURCE_HPP
+#define BOOST_JSON_DETAIL_MONOTONIC_RESOURCE_HPP
+
+#include <boost/json/detail/config.hpp>
+
+#include <cstddef>
+
+namespace boost {
+namespace json {
+namespace detail {
+namespace empty_types {
+
+struct class_type {};
+enum unscoped_enumeration_type { };
+enum class scoped_enumeration_type { };
+
+} // empty_types
+
+// [support.types] p5: The type max_­align_­t is a trivial
+// standard-layout type whose alignment requirement
+// is at least as great as that of every scalar type.
+struct max_align_t
+{
+    // arithmetic types
+    char a_;
+    char16_t b_;
+    char32_t c_;
+    bool d_;
+    short int e_;
+    int f_;
+    long int g_;
+    long long int h_;
+    wchar_t i_;
+    float j_;
+    double k_;
+    long double l_;
+    // enumeration types
+    empty_types::unscoped_enumeration_type m_;
+    empty_types::scoped_enumeration_type n_;
+    // pointer types
+    void* o_;
+    char* p_;
+    empty_types::class_type* q_;
+    empty_types::unscoped_enumeration_type* r_;
+    empty_types::scoped_enumeration_type* s_;
+    void(*t_)();
+    // pointer to member types
+    char empty_types::class_type::* u_;
+    void (empty_types::class_type::*v_)();
+    // nullptr
+    std::nullptr_t w_;
+};
+
+inline
+constexpr
+std::size_t
+max_align()
+{
+#ifdef BOOST_JSON_STANDALONE
+    return alignof(std::max_align_t);
+#else
+    return alignof(max_align_t);
+#endif
+}
+
+inline
+unsigned char*
+align_up(
+    unsigned char* ptr,
+    std::size_t align)
+{
+    // alignment shall be a power of two
+    BOOST_ASSERT(!(align & (align - 1)));
+    return reinterpret_cast<unsigned char*>(
+        (reinterpret_cast<std::uintptr_t>(ptr) + align - 1) & ~(align - 1));
+}
+
+} // detail
+} // json
+} // boost
+
+#endif
--- a/include/boost/json/impl/monotonic_resource.ipp
+++ b/include/boost/json/impl/monotonic_resource.ipp
@@ -1,5 +1,6 @@
 //
 // Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+// Copyright (c) 2020 Krystian Stasiowski (sdkrystian@gmail.com)
 //
 // Distributed under the Boost Software License, Version 1.0. (See accompanying
 // file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
@@ -12,59 +13,109 @@
 
 #include <boost/json/monotonic_resource.hpp>
 
+#include <memory>
+
 namespace boost {
 namespace json {
 
-struct monotonic_resource::block
+// ensures that the alignment of base is
+// the strictest fundamental alignment requirement
+struct alignas(detail::max_align())
+    monotonic_resource::block
 {
-    std::size_t const size;
-    std::uintptr_t top;
-    block* next;
+    std::size_t size;
+    block* next = nullptr;
+    unsigned char* base;
+    unsigned char* top;
 
     block(
-        std::size_t size_,
-        block* next_)
-        : size(size_)
-        , top(reinterpret_cast<
-            std::uintptr_t>(this+1))
-        , next(next_)
-    {
-    }
-
-    void*
-    alloc(
+        unsigned char* data,
         std::size_t n,
-        std::size_t align) noexcept
+        block* head)
+        : size(n)
+        , next(head)
+        , base(data)
+        , top(data)
     {
-        // must be power of 2
-        BOOST_ASSERT(
-            (align & (align - 1)) == 0);
-        auto i = top;
-        if((i & (align - 1)) != 0)
-            i = (i | (align - 1)) + 1;
-        if(i + n > (reinterpret_cast<
-            std::uintptr_t>(this+1) + size))
-            return nullptr;
-        top = i + n;
-        return reinterpret_cast<void*>(i);
     }
 };
 
+template<typename Block>
+void*
+monotonic_resource::
+allocate_in_block(
+    Block& b,
+    std::size_t size,
+    std::size_t align)
+{
+    const auto aligned = 
+        detail::align_up(b.top, align);
+    const auto next_top = aligned + size;
+    if(next_top <= b.base + b.size)
+    {
+        b.top = next_top;
+        return aligned;
+    }
+    return nullptr;
+}
+
 auto
 monotonic_resource::
-alloc_block(std::size_t size) ->
+allocate_new_block(std::size_t size) ->
     block&
+{  
+    /*-------------------------------------  
+             block storage layout:        
+           [ n ] = n allocated bytes
+      
+                header         data
+           [ sizeof(block) ] [ size ]
+      
+        data is guaranteed to be aligned
+            to alignof(max_align_t)
+    -------------------------------------*/
+    const auto bytes = size + sizeof(block);
+    const auto data = static_cast<unsigned char*>(
+        ::operator new(bytes));
+    head_ = ::new(static_cast<void*>(data))
+        block(data + sizeof(block), size, head_);
+    return *head_;
+}
+
+// returns the closest power of two
+// that is greater than requested
+std::size_t
+monotonic_resource::
+next_block_size(std::size_t requested)
 {
-    auto const n = (
-        size + sizeof(block) - 1) /
-        sizeof(block);
-    auto const bytes =
-        (n + 1) * sizeof(block);
-    auto& b = *::new(
-        ::operator new(bytes)) block(
-            n * sizeof(block), head_);
-    head_ = &b;
-    return b;
+    if(requested >= max_block_size_ >> 1)
+        return max_block_size_;
+    if(requested <= min_block_size_)
+        return min_block_size_;
+    std::size_t next = 1;
+    for(; requested; requested >>= 1, next <<= 1);
+    return next;
+}
+
+// skips rounding if requested
+// is already a power of two
+std::size_t
+monotonic_resource::
+closest_block_size(std::size_t requested)
+{
+    if(requested & (requested - 1))
+        return next_block_size(requested);
+    return requested;
+}
+
+std::size_t
+monotonic_resource::
+grow_block_size(std::size_t size)
+{
+    // prevents overflow
+    if(size >= max_block_size_ >> 1)
+        return max_block_size_;
+    return size << 1;
 }
 
 monotonic_resource::
@@ -79,15 +130,18 @@ monotonic_resource::
 }
 
 monotonic_resource::
-monotonic_resource() noexcept
-    : block_size_(64 * 1024)
+monotonic_resource(
+    std::size_t initial_size) noexcept
+    : block_size_(closest_block_size(initial_size))
 {
 }
 
 monotonic_resource::
 monotonic_resource(
-    std::size_t block_size) noexcept
-    : block_size_(block_size)
+    void* buffer,
+    std::size_t buffer_size) noexcept
+    : block_size_(next_block_size(buffer_size))
+    , initial_(buffer, buffer_size)
 {
 }
 
@@ -97,17 +151,18 @@ do_allocate(
     std::size_t n,
     std::size_t align)
 {
-    if(head_)
-    {
-        auto p = head_->alloc(n, align);
-        if(p)
+    if(initial_.base)
+        if(auto p = allocate_in_block(initial_, n, align))
             return p;
-    }
-    if(n > block_size_ - 2 * align)
-        alloc_block(n + 2 * align);
-    else
-        alloc_block(block_size_);
-    auto p = head_->alloc(n, align);
+    if(head_)
+        if(auto p = allocate_in_block(*head_, n, align))
+            return p;
+    auto new_block_size = block_size_;
+    if(n > block_size_)
+        new_block_size = closest_block_size(n);
+    auto p = allocate_in_block(
+        allocate_new_block(new_block_size), n, align);
+    block_size_ = grow_block_size(new_block_size);
     BOOST_ASSERT(p);
     return p;
 }
@@ -119,7 +174,7 @@ do_deallocate(
     std::size_t,
     std::size_t)
 {
-    // nothing
+    // do nothing
 }
 
 bool
--- a/include/boost/json/monotonic_resource.hpp
+++ b/include/boost/json/monotonic_resource.hpp
@@ -1,5 +1,6 @@
 //
 // Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+// Copyright (c) 2020 Krystian Stasiowski (sdkrystian@gmail.com)
 //
 // Distributed under the Boost Software License, Version 1.0. (See accompanying
 // file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
@@ -12,6 +13,7 @@
 
 #include <boost/json/detail/config.hpp>
 #include <boost/json/storage_ptr.hpp>
+#include <boost/json/detail/monotonic_resource.hpp>
 
 namespace boost {
 namespace json {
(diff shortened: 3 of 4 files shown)


vinniefalco:
The commit message is too long, since this is pre-release we can just make it "Refactor monotonic_resource."

vinniefalco:
Tests are missing.

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -79,35 +97,67 @@ monotonic_resource::
 }
 
 monotonic_resource::
-monotonic_resource() noexcept
-    : block_size_(64 * 1024)
+monotonic_resource(
+    std::size_t block_size,
```
vinniefalco, line 101: I think "block_size" is the wrong parameter. We should track the standard here. Instead of block_size we should make it `initial_size` and let that be the size of the first block allocated. And we should switch around the parameters, `(max_alloc, initial_size)` and provide defaults for both of them, with `initial_size` defaulting to zero (meaning, to use some internal constant). Again we should track the standard here, and the standard doesn't even have `initial_size` so there's that.

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -79,35 +97,67 @@ monotonic_resource::
...
+    : block_size_(block_size),
+      max_alloc_(max_alloc)
 {
 }
 
 monotonic_resource::
 monotonic_resource(
-    std::size_t block_size) noexcept
-    : block_size_(block_size)
+  void* initial_block,
```
vinniefalco, line 110: This parameter has the wrong name, it isn't a "block" in the sense of being prefixed with a `block` in the layout

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -18,38 +19,55 @@ namespace json {
...
+        , top(base_ptr())
     {
     }
 
-    void*
-    alloc(
-        std::size_t n,
-        std::size_t align) noexcept
+    std::uintptr_t
+    base_ptr() const
```
vinniefalco, line 35: or `data() noexcept const`

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -18,38 +19,55 @@ namespace json {
...
     }
 };
 
+std::uintptr_t
+monotonic_resource::
+align_top(
+    std::uintptr_t ptr,
+    std::size_t align)
+{
+    // must be power of 2
```
vinniefalco, line 48: https://cplusplus.github.io/LWG/issue2843

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -18,38 +19,55 @@ namespace json {
...
+    auto i = ptr;
+    return i & (align - 1) ?
+        (i | (align - 1)) + 1 : i;
+}
+
+template<typename Block>
+void*
+monotonic_resource::
+alloc_in_block(
+    Block& blk,
```
vinniefalco, line 60: or just `b` - nothing wrong with single letter variables, we use them all the time in math.

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -18,38 +19,55 @@ namespace json {
...
+        (align & (align - 1)) == 0);
+    auto i = ptr;
+    return i & (align - 1) ?
+        (i | (align - 1)) + 1 : i;
+}
+
+template<typename Block>
+void*
+monotonic_resource::
+alloc_in_block(
```
vinniefalco, line 59: The "name" of a function includes the types of its parameters, so `alloc( Block& b, std::size_t n, std::size_t align )` says the same thing as `alloc_in_block( ...`

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -79,35 +97,69 @@ monotonic_resource::
...
 {
 }
 
 monotonic_resource::
 monotonic_resource(
-    std::size_t block_size) noexcept
-    : block_size_(block_size)
+  void* buffer,
+  std::size_t buffer_size,
+  std::size_t max_alloc) noexcept
```
vinniefalco, line 112: The indentation is off here

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -79,35 +97,69 @@ monotonic_resource::
...
     std::size_t n,
     std::size_t align)
 {
+    // use the initial block if possible
+    if(initial_.base)
+        if (auto p = alloc_in_block(initial_, n, align))
+            return p;
+    // cap the number of dynamically allocated bytes
+    if(allocated_ + n > max_alloc_)
+        throw std::bad_alloc();
```
vinniefalco, line 149: This needs to be like `detail::raise_bad_alloc` and then in that function use `BOOST_THROW_EXCEPTION`, see:
https://github.com/CPPAlliance/json/blob/4a6a6e7cc3f02f6980e68557b9843095a578934c/include/boost/json/impl/except.ipp#L42

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -79,35 +97,69 @@ monotonic_resource::
...
 monotonic_resource::
 do_allocate(
     std::size_t n,
     std::size_t align)
 {
+    // use the initial block if possible
+    if(initial_.base)
+        if (auto p = alloc_in_block(initial_, n, align))
+            return p;
+    // cap the number of dynamically allocated bytes
```
vinniefalco, line 147: This comment adds no information

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -79,35 +97,69 @@ monotonic_resource::
...
+    if(allocated_ + n > max_alloc_)
+        throw std::bad_alloc();
+    // do this after the check for exception guarantee
```
vinniefalco, line 150: Tip, explain intent using the word "otherwise". Example:
```
// Increment after the throw,
// otherwise no strong exception safety.
```

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -79,35 +97,69 @@ monotonic_resource::
...
+        if(auto p = alloc_in_block(*head_, n, align))
             return p;
-    }
-    if(n > block_size_ - 2 * align)
-        alloc_block(n + 2 * align);
-    else
-        alloc_block(block_size_);
-    auto p = head_->alloc(n, align);
+    // if the requested size is greater than the block size,
+    // use the requested size instead
```
vinniefalco, line 156: This comment adds no information

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -79,35 +97,69 @@ monotonic_resource::
...
+    const auto new_block_size = n > block_size_ - 2 * align ?
+        n + 2 * align : block_size_;
```
vinniefalco, line 158: Best not the use the ternary operator here, because of code coverage. It isn't quite as tidy but this is better:
```
std::size_t new_size;
if( n > block_size_ - 2 * align )
  new_size = n + 2 * align;
else
  new_size = block_size_;
```

Also are we checking that `n + 2 * align` does not overflow?

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -79,35 +97,69 @@ monotonic_resource::
...
+    alloc_block(new_block_size);
+    // apply the growth factor for the next block
```
vinniefalco, line 160: LOL useless comment...

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -79,35 +97,69 @@ monotonic_resource::
...
+    block_size_ = new_block_size * 2;
```
vinniefalco, line 161: This can overflow

include/boost/json/monotonic_resource.hpp
```diff
@@ -20,23 +21,79 @@ namespace json {
...
     : public memory_resource
-{
+{   
+    struct initial_block
+    {
+        std::size_t size;
+        void* base = nullptr;
+        std::uintptr_t top;
+
+        initial_block() { }
```
vinniefalco, line 31: why not default this?

include/boost/json/monotonic_resource.hpp
```diff
@@ -20,23 +21,79 @@ namespace json {
...
+            void* data_,
+            std::size_t size_)
+            : size(size_)
+            , base(data_)
+            , top(base_ptr())
+        {
+        }
+
+        std::uintptr_t
+        base_ptr() const
```
vinniefalco, line 43: `data()`

include/boost/json/monotonic_resource.hpp
```diff
@@ -20,23 +21,79 @@ namespace json {
...
+        }
+    };
+
     struct block;
-    std::size_t const block_size_ = 64 * 1024;
-    block* head_ = nullptr;
 
+    std::size_t block_size_;
+    block* head_ = nullptr;
+    // initial unmanaged block
```
vinniefalco, line 54: Useless comment

include/boost/json/monotonic_resource.hpp
```diff
@@ -20,23 +21,79 @@ namespace json {
...
+    initial_block initial_;
+    // maximum number of bytes that
+    // can be dynamically allocated
```
vinniefalco, line 57: Useless comment

include/boost/json/monotonic_resource.hpp
```diff
@@ -20,23 +21,79 @@ namespace json {
...
+    std::size_t const max_alloc_;
+    std::size_t allocated_ = 0;
+    
     inline block& alloc_block(std::size_t size);
+    
+    template<typename Block>
+    void*
+    alloc_in_block(
+        Block& blk,
```
vinniefalco, line 66: Nice, recycle the function lol

include/boost/json/impl/monotonic_resource.ipp
sdkrystian: Would an IILE work?

include/boost/json/impl/monotonic_resource.ipp
sdkrystian: Is there a point to doing this? It's implementation detail. `data()` usually implies that you get a pointer; this returns an integer.

include/boost/json/impl/monotonic_resource.ipp
sdkrystian: Copy

include/boost/json/impl/monotonic_resource.ipp
vinniefalco: oh.... true. You called it `base_ptr` though :) maybe `base()`?

include/boost/json/monotonic_resource.hpp
sdkrystian: More characters for identical initialization semantics (albeit defaulted would make the struct trivial). I'll change it.

include/boost/json/impl/monotonic_resource.ipp
sdkrystian: Affirm

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -10,61 +11,110 @@
 #ifndef BOOST_JSON_IMPL_MONOTONIC_RESOURCE_IPP
 #define BOOST_JSON_IMPL_MONOTONIC_RESOURCE_IPP
 
+#include <algorithm>
```
vinniefalco, line 14: std includes should come last. Rationale: Placing them last makes it more likely for a compiler error to occur if an include is missing.

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -10,61 +11,110 @@
...
+
 #include <boost/json/monotonic_resource.hpp>
 
 namespace boost {
 namespace json {
 
-struct monotonic_resource::block
+// ensures that the alignment of top is
+// the strictest fundamental alignment requirement
+struct alignas(alignof(std::max_align_t))
```
vinniefalco, line 23: Is this C++11?

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -10,61 +11,110 @@
...
+    const auto bytes =
+      size + sizeof(block);
+    head_ = ::new(::operator new(bytes))
+        block(size, head_);
+    return *head_;
+}
+
+std::size_t
+monotonic_resource::
+guess_next_size(std::size_t minimum)
```
vinniefalco, line 97: it isn't a guess though, it is precise

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -97,17 +153,24 @@ do_allocate(
...
-    else
-        alloc_block(block_size_);
-    auto p = head_->alloc(n, align);
+    // neither of the above was sucessful,
+    // so we need to allocate a new block.
+    // if the requested size is greater than the block size,
+    // use the requested size plus block overhead rounded up to
+    // the closest power of 2
+    auto new_block_size = n > block_size_ ?
+        guess_next_size(n) : block_size_;
```
vinniefalco, line 170: Please use `if` `else` statements instead of `?`...

include/boost/json/impl/monotonic_resource.ipp
vinniefalco: ugh too try-hard

include/boost/json/impl/monotonic_resource.ipp
vinniefalco: why do we need `<algorithm>` that's an expensive header

include/boost/json/impl/monotonic_resource.ipp
sdkrystian: https://en.cppreference.com/w/cpp/language/alignas

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -10,61 +11,113 @@
...
+void*
+monotonic_resource::
+allocate_in_block(
+    Block& blk,
+    std::size_t size,
+    std::size_t align)
+{
+    auto aligned_addr =
+        align_address(blk.top, align);
+    if(aligned_addr + size > blk.base() + blk.size)
```
vinniefalco, line 70: you should structure the LIKELY case to be true. This condition will be false more often than it is true, thus you should invert it:
```
if(aligned_addr + size <= blk.base() + blk.size)
{
    blk.top = aligned_addr + size;
    return reinterpret_cast<void*>(aligned_addr);
}
return nullptr;
```

include/boost/json/impl/monotonic_resource.ipp
sdkrystian: `std::max`

vinniefalco: its a shame to have to include that just for one piss-ant function

test/monotonic_resource.cpp
```diff
@@ -14,12 +15,108 @@
 
 #include "test_suite.hpp"
 
+#include <memory>
+#include <iostream>
+
 namespace boost {
 namespace json {
 
 class pool_test
```
vinniefalco, line 24: This should say `monotonic_resource_test`

test/monotonic_resource.cpp
```diff
@@ -14,12 +15,108 @@
...
+        // test if sizes are correctly determined from initial buffers
+        {
+            {   
+                std::unique_ptr<unsigned char[]> buf(new unsigned char[512]);
+                monotonic_resource mr(buf.get(), 512);
+                BOOST_TEST(all_alloc_in_same_block(mr, 512, 1));
+                BOOST_TEST(all_alloc_in_same_block(mr, 1024, 1));
+            }
+            {   
+                std::unique_ptr<unsigned char[]> buf(new unsigned char[2048]);
```
vinniefalco, line 106: why not make this a local array, e.g.
```
char buf[2048];
```
? That's going to be the most common use-case.

include/boost/json/monotonic_resource.hpp
```diff
@@ -13,30 +14,88 @@
...
+
+    template<typename Block>
+    void*
+    allocate_in_block(
+        Block& blk,
+        std::size_t n,
+        std::size_t align);
+
+    block&
+    allocate_new_block(std::size_t size);
```
vinniefalco, line 75: this needs to be marked `inline` no?

vinniefalco:
Warnings and errors: https://travis-ci.org/github/CPPAlliance/json

test/monotonic_resource.cpp
sdkrystian: I don't think it really matters here... I can add test for both.

test/monotonic_resource.cpp
sdkrystian: Copy

include/boost/json/monotonic_resource.hpp
sdkrystian: Ah, forgot about header only. Should just be inline though since these aren't part of the interface, right?

test/monotonic_resource.cpp
vinniefalco: its just that now we're instantiating `unique_ptr` needlessly

include/boost/json/impl/monotonic_resource.ipp
sdkrystian: It's gone now

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -15,56 +16,124 @@
...
     block(
         std::size_t size_,
         block* next_)
-        : size(size_)
-        , top(reinterpret_cast<
-            std::uintptr_t>(this+1))
-        , next(next_)
+        : size(size_), 
+          next(next_),
+          top(base())
```
vinniefalco, line 33:
```
        : size(size_)
        , next(next_)
        , top(base())
```
This makes it easier to reorder them

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -12,59 +13,116 @@
...
-        sizeof(block);
-    auto const bytes =
-        (n + 1) * sizeof(block);
-    auto& b = *::new(
-        ::operator new(bytes)) block(
-            n * sizeof(block), head_);
-    head_ = &b;
-    return b;
+    // returns the closest power of two
+    // that is greater than requested
```
vinniefalco, line 95: This comment goes on the function not inside the body

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -12,59 +13,116 @@
...
+    std::size_t next = 1;
+    for(; requested; requested >>= 1, next <<= 1);
+    return next;
+}
+
+std::size_t
+monotonic_resource::
+next_block_size(
+    std::size_t requested,
+    std::false_type)
```
vinniefalco, line 109: What the hell is this true_type false_type? I have no idea. Couldn't we just name the functions differently?

include/boost/json/monotonic_resource.hpp
```diff
@@ -13,30 +14,97 @@
...
 #include <boost/json/storage_ptr.hpp>
 
+#include <limits>
+#include <new>
+
+#ifndef BOOST_JSON_MAX_ALIGN
+# ifdef BOOST_JSON_STANDALONE
+#  define BOOST_JSON_MAX_ALIGN alignof(std::max_align_t)
+# else
+#  define BOOST_JSON_MAX_ALIGN boost::move_detail::alignment_of<boost::move_detail::max_align_t>::value
```
vinniefalco, line 24: This adds a dependency on Boost.Move? Why is it accessing a detail namespace? That's bad form.

include/boost/json/monotonic_resource.hpp
```diff
@@ -13,30 +14,97 @@
...
+
+    struct alignas(BOOST_JSON_MAX_ALIGN) block;
+
+    std::size_t block_size_;
     block* head_ = nullptr;
+    initial_block initial_;
+    
+    static const std::size_t min_block_size_ = 1024;
+    static const std::size_t max_block_size_ = 
+        std::numeric_limits<std::size_t>::max();
```
vinniefalco, line 62: You could avoid this numeric_limits include by simply writing `std::size_t(-1)`

include/boost/json/monotonic_resource.hpp
```diff
@@ -13,30 +14,97 @@
...
+
+    inline
+    std::size_t
+    grow_buffer(std::size_t size);
 
 public:
     BOOST_JSON_DECL
     ~monotonic_resource();
 
     BOOST_JSON_DECL
```
vinniefalco, line 91: Why are we putting the `BOOST_JSON_DECL` modified on deleted functions?

include/boost/json/detail/monotonic_resource.hpp
```diff
@@ -0,0 +1,91 @@
...
+};
+
+inline
+constexpr
+std::size_t
+max_align()
+{
+    return alignof(
+#ifdef BOOST_JSON_STANDALONE
+        std::
```
vinniefalco, line 70: jesus H this is ugly...never split an identifier's scope with a macro! Use
```
#ifdef BOOST_JSON_STANDALONE
        std::max_align_t
#else
       max_align_t
#endif
```

include/boost/json/detail/monotonic_resource.hpp
```diff
@@ -0,0 +1,91 @@
...
+#define BOOST_JSON_DETAIL_MONOTONIC_RESOURCE_HPP
+
+#include <boost/json/detail/config.hpp>
+
+#include <cstddef>
+
+namespace boost {
+namespace json {
+namespace detail {
+namespace dummy {
```
vinniefalco, line 20: I don't like the word "dummy" appearing in my source code. Use `other` or something sterile.

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -12,59 +13,109 @@
...
+closest_block_size(std::size_t requested)
+{
+    if(requested & (requested - 1))
+        return next_block_size(requested);
+    return requested;
+}
+
+std::size_t
+monotonic_resource::
+grow_block_size(std::size_t size)
```
vinniefalco, line 113: This is 112% better than overloads with `true_type` and `false_type` !!

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -97,17 +151,25 @@ do_allocate(
...
+        if(auto p = allocate_in_block(*head_, n, align))
             return p;
-    }
-    if(n > block_size_ - 2 * align)
-        alloc_block(n + 2 * align);
-    else
-        alloc_block(block_size_);
-    auto p = head_->alloc(n, align);
+    // neither of the above was sucessful,
+    // so we need to allocate a new block.
```
vinniefalco, line 163: This comment adds no information

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -97,17 +151,25 @@ do_allocate(
...
+    // if the requested size is greater than the block size,
+    // use the requested size plus block overhead rounded up to
+    // the closest power of 2
```
vinniefalco, line 166: This comment also adds no information

include/boost/json/impl/monotonic_resource.ipp
sdkrystian: Those comments just walk through the process -- makes it much faster to understand what's going on

include/boost/json/impl/monotonic_resource.ipp
sdkrystian: Epic

include/boost/json/detail/monotonic_resource.hpp
sdkrystian: Hah, was being clever. I'll change it.

include/boost/json/detail/monotonic_resource.hpp
sdkrystian: How about `empty_types`

vinniefalco: what's wrong with `other` ?

sdkrystian: It provides no information as to what it contains

PR #91, opened 2020-05-14
monotonic_resource docs

sdkrystian:


include/boost/json/monotonic_resource.hpp
```diff
@@ -18,11 +18,37 @@
 namespace boost {
 namespace json {
 
-/** A fast memory resource using many large fixed size blocks.
+/** A fast memory resource that never deallocates.
+    
+    `monotonic_resource` allocates large fixed-size
```
vinniefalco, line 23: Avoid repeating the type

include/boost/json/monotonic_resource.hpp
```diff
@@ -18,11 +18,37 @@
...
+    blocks of memory, with each allocated block being
+    twice as large as the previous block.
+    Allocated blocks are not freed until the memory
+    resource is destroyed; calls to @ref do_deallocate
+    have no effect. Memory will not be allocated unless
+    @ref do_allocate is called and the current block
```
vinniefalco, line 29: The user doesn't call `do_allocate` so this is confusing. The user calls `memory_resource::allocate`.

include/boost/json/monotonic_resource.hpp
```diff
@@ -72,36 +98,126 @@ class monotonic_resource final
...
+
+        @return A pointer to memory that is of size `n` and
+        suitably aligned for `align`.
+
+        @param n The number of bytes to allocate.
+
+        @param align The desired alignment of the memory.
+
+        @throw std::bad_alloc Failure to allocate a new block.
+    */
```
vinniefalco, line 194: This is not going to work out well, we need to discuss it

include/boost/json/monotonic_resource.hpp
```diff
@@ -18,11 +18,37 @@
...
+    at which point subsequent blocks will be allocated
+    using the default resource.
+
+    In addition to this, an *initial size* can be
+    passed to the constructor of `monotonic_resource`.
+    The initial size determines the size of the first
+    dynamically allocated block.
+
+    This memory resource is useful for parsing without
+    subsequent modification to the resulting #ref value.
```
vinniefalco, line 46: `@ref`

include/boost/json/monotonic_resource.hpp
```diff
@@ -18,11 +18,37 @@
...
 */
 class monotonic_resource final
     : public memory_resource
 {   
+#ifndef BOOST_JSON_DOCS
```
vinniefalco, line 51: shouldn't need this but check the docs

include/boost/json/monotonic_resource.hpp
sdkrystian: The way the exception is specified? Or is it because it doesn't directly throw the exception?

I can just add a not saying that it calls `::operator new` and that might throw.

vinniefalco: Because of how it is derived from memory_resource - we talked about this already.

sdkrystian: Ah, we resolved that already then.

PR #112, opened 2020-06-14
Make key_value_pair tuple-like

### LINKED ISSUE #111: tuple-like key_value_pair (for structured bindings)
At the moment `key_value_pair` cannot be used for structured bindings without user boilerplate.
Since it does model a pair anyway and offers an interface similar to `std::pair`, it would be nice for it to also specialize `std::tuple_size`, `std::tuple_element` and a form of `get` when using C++17 or above
This would make the following code well-formed in C++17 or above:
```cpp
object o{{"one", 1}, {"two", 2}};
const auto it = o.find("one");
if (it == o.end())
  handle_error();
const auto& [key, value] = *it;
```
Note that replacing `object` with `std::map<std::string, int>` would produce well-formed C++17 code (see https://godbolt.org/z/fVZUkC)

If this is something the authors would like too see too, I'd happily make a PR for it

AeroStun:
Fixes  #111 
Reviews and other comments are welcome

### DIFF
--- a/include/boost/json/value.hpp
+++ b/include/boost/json/value.hpp
@@ -2830,8 +2830,134 @@ private:
     std::uint32_t next_;
 };
 
+//----------------------------------------------------------
+
+template <std::size_t I>
+auto
+get(key_value_pair const&) noexcept ->
+        typename std::conditional<I == 0,
+            string_view const,
+            json::value const&
+        >::type
+{
+    BOOST_STATIC_ASSERT(I == -1u);
+}
+
+template <std::size_t I>
+auto
+get(key_value_pair&) noexcept ->
+        typename std::conditional<I == 0,
+            string_view const,
+            json::value&
+        >::type
+{
+    BOOST_STATIC_ASSERT(I == -1u);
+}
+
+template <std::size_t I>
+auto
+get(key_value_pair&&) noexcept ->
+typename std::conditional<I == 0,
+    string_view const,
+    json::value&&
+>::type
+{
+    BOOST_STATIC_ASSERT(I == -1u);
+}
+
+/** Extracts a key_value_pair's key using tuple-like interface
+*/
+template <>
+inline
+string_view const
+get<0>(key_value_pair const& kvp) noexcept {
+    return kvp.key();
+}
+
+/** Extracts a key_value_pair's key using tuple-like interface
+*/
+template <>
+inline
+string_view const
+get<0>(key_value_pair& kvp) noexcept
+{
+  return kvp.key();
+}
+
+/** Extracts a key_value_pair's key using tuple-like interface
+*/
+template <>
+inline
+string_view const
+get<0>(key_value_pair&& kvp) noexcept
+{
+  return kvp.key();
+}
+
+/** Extracts a key_value_pair's value using tuple-like interface
+*/
+template <>
+inline
+json::value const&
+get<1>(key_value_pair const& kvp) noexcept
+{
+  return kvp.value();
+}
+
+/** Extracts a key_value_pair's value using tuple-like interface
+*/
+template <>
+inline
+json::value&
+get<1>(key_value_pair& kvp) noexcept
+{
+  return kvp.value();
+}
+
+/** Extracts a key_value_pair's value using tuple-like interface
+*/
+template <>
+inline
+json::value&&
+get<1>(key_value_pair&& kvp) noexcept
+{
+  return std::move(kvp.value());
+}
+
 } // json
 } // boost
+
+namespace std
+{
+
+/** Tuple-like size access for key_value_pair
+*/
+template <>
+struct tuple_size<::boost::json::key_value_pair>
+    : std::integral_constant<std::size_t, 2> {};
+
+/** Tuple-like access for the key type of key_value_pair
+*/
+template <>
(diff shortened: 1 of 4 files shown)


test/value.cpp
```diff
@@ -1583,6 +1583,38 @@ class value_test
             v3.value().get_string() == "value");
         BOOST_TEST(std::memcmp(
             v3.key_c_str(), "key\0", 4) == 0);
+
+        BOOST_STATIC_ASSERT(std::tuple_size<key_value_pair>::value == 2);
+        BOOST_TEST(get<0>(v2) == "key");
+        BOOST_TEST(
+            get<1>(v2).get_string() == "value");
+
+#if __cplusplus >= 201703L
```
madmongo1, line 1592: This won't fly on MSVC

include/boost/json/value.hpp
```diff
@@ -2830,9 +2830,104 @@ struct key_value_pair
...
+*/
+template <>
+struct tuple_element<1, ::boost::json::key_value_pair> {
+  using type = ::boost::json::value&;
+};
+
+/** Tuple-like access for the value type of key_value_pair
+*/
+template <>
+struct tuple_element<1, ::boost::json::key_value_pair const> {
```
madmongo1, line 2925: No r-value overloads

test/value.cpp
AeroStun: Good to know! Thanks for pointing it out; I believe I have now fixed that

test/value.cpp
```diff
@@ -1589,7 +1589,7 @@ class value_test
         BOOST_TEST(
             get<1>(v2).get_string() == "value");
 
-#if __cplusplus >= 201703L
+#if __cplusplus >= 201703L || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
         auto const [kc, vc] = v1;
```
vinniefalco, line 1593: does `vc` receive a copy of the `json::value`?

AeroStun: Correct; is not a desired behavior?

vinniefalco: I haven't used structured bindings, so I don't know. What if you want a reference? do you write  `auto const& [kc, kv] = v1`? Or do you write `auto const [kc, &kv] = v1]` ?

AeroStun: `auto const& [kc, kv]`
See the various `static_assert`s in the added test code
I'm still struggling to get rvalues to work tho; I've actually fixed Richard's second comment, but didn't close it yet because I now want `auto [k, v] = kvp_prvalue` to be well formed and move the value, and likewise with `auto&& [krr, vrr] = kvp_prvalue`
Been staring at [\[dcl.struct.bind\]#4](https://eel.is/c++draft/dcl.struct.bind#4) and I'm still unsure as to why I can't both lvalues and rvalues to work without copying when `auto` and `auto&&` are used.

vinniefalco: `auto const& [kc, kv]` is interesting though, do you get a reference to a temporary `string_view`?

AeroStun: No; the `string_view` is copied
```cpp
auto const& [kcr, vcr] = v1;
static_assert(std::is_same_v<decltype(kcr), string_view const>);
static_assert(std::is_same_v<decltype(vcr), value const&>);
```

include/boost/json/value.hpp
```diff
@@ -2854,6 +2854,17 @@ get(key_value_pair&) noexcept ->
...
 
+template <std::size_t I>
+auto
+get(key_value_pair&&) noexcept ->
+typename std::conditional<I == 0,
+    string_view const,
+    json::value&&
+>::type
+{
+  BOOST_STATIC_ASSERT(I == -1u);
```
vinniefalco, line 2865: nit: we use 4 spaces per tab

test/value_from.cpp
```diff
@@ -183,6 +186,13 @@ class value_from_test
             BOOST_TEST(c.is_array());
             BOOST_TEST(to_string(c) == to_string(b));
         }
+        {
+            key_value_pair a{"2", 1};
+            value b{"2", 1};
+            value c = value_from(a);
+            BOOST_TEST(c.is_array());
+            BOOST_TEST(to_string(c) == to_string(b));
+        }
```
AeroStun, line 195: @sdkrystian pointed out to me that `value_from` has special support for tuple-like types, hence I added this
However I am not a 100% sure whether `value_from(key_value_pair)` should yield an object with a single field, or an array with two elements (the current default).

sdkrystian: Two element array is correct.

AeroStun:
(squashed)

include/boost/json/value.hpp
```diff
@@ -2830,9 +2830,135 @@ struct key_value_pair
     std::uint32_t next_;
 };
 
+//----------------------------------------------------------
+
+template <std::size_t I>
+auto
+get(key_value_pair const&) noexcept ->
```
vinniefalco, line 2837: These are missing Javadocs. I think it would be best to ifdef all of this so it is invisible to the documentation tool, and put a generic declaration in there just for the docs. Something like this:
```
#ifdef BOOST_JSON_DOCS

/** Tuple-like interface
*/
template< std::size_t I, class T >
__see_below__
get( T&& );

#else

// actual defintions

#endif
```

PR #114, opened 2020-06-20
Flush buffer after parsing escapes

### LINKED ISSUE #113: valid json reported as invalid
The file below contains a valid json string. However, the boost json example program `validate` reports it is invalid, and the example program `pretty` reports `caught exception: extra data`. 

[valid_json_boost_json_issue.txt](https://github.com/CPPAlliance/json/files/4803279/valid_json_boost_json_issue.txt)

Many small changes to the string makes it parse correctly, such as deleting an innocuous looking character from the string.

I've reproduced this on the tip of master (f47fbeb1a54b832221a35c17912febb7a2c3a317) with both clang and gcc.

Please let me know any additional information I can provide that would be helpful.

sdkrystian:
Fixes #113

### DIFF
--- a/include/boost/json/basic_parser.hpp
+++ b/include/boost/json/basic_parser.hpp
@@ -734,22 +734,25 @@ parse_string(
         start = cs.data();
         state st;
         st_.pop(st);
-        switch(st)
+        if (st != state::str1)
         {
-        default:
-        case state::str1: goto do_str1;
-        case state::str2: goto do_str2;
-        case state::str3: goto do_str3;
-        case state::str4: goto do_str4;
-        case state::str5: goto do_str5;
-        case state::str6: goto do_str6;
-        case state::str7: goto do_str7;
-        case state::sur1: goto do_sur1;
-        case state::sur2: goto do_sur2;
-        case state::sur3: goto do_sur3;
-        case state::sur4: goto do_sur4;
-        case state::sur5: goto do_sur5;
-        case state::sur6: goto do_sur6;
+            cs.clip(temp.max_size());
+            switch(st)
+            {
+            default:
+            case state::str2: goto do_str2;
+            case state::str3: goto do_str3;
+            case state::str4: goto do_str4;
+            case state::str5: goto do_str5;
+            case state::str6: goto do_str6;
+            case state::str7: goto do_str7;
+            case state::sur1: goto do_sur1;
+            case state::sur2: goto do_sur2;
+            case state::sur3: goto do_sur3;
+            case state::sur4: goto do_sur4;
+            case state::sur5: goto do_sur5;
+            case state::sur6: goto do_sur6;
+            }
         }
     }
 
@@ -757,7 +760,6 @@ parse_string(
     //
     // zero-copy unescaped runs
     //
-do_str1:
     cs.skip(detail::count_unescaped(
         cs.data(), cs.remain()));
     for(;;)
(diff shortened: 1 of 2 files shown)


include/boost/json/basic_parser.hpp
```diff
@@ -734,30 +734,32 @@ parse_string(
         start = cs.data();
         state st;
         st_.pop(st);
-        switch(st)
+        if (st != state::str1)
```
vinniefalco, line 737: This is now irregular (different from all the other switch-restore stanzas in the file).

include/boost/json/basic_parser.hpp
```diff
@@ -807,6 +809,13 @@ parse_string(
                             return result::fail;
                     }
                 }
+                // KRYSTIAN TODO: this check is performed again
+                // at the label. Clip & increment the stream, then jump to
+                // do_str3
+
+                // JSON escapes can never make the
+                // transcoded utf8 string larger.
+                cs.clip(temp.max_size());
```
vinniefalco, line 818: Why `max_size` instead of `capacity`?

include/boost/json/basic_parser.hpp
```diff
@@ -1094,30 +1100,33 @@ parse_string(
             ++cs;
             continue;
         }
-        if(BOOST_JSON_LIKELY(more_))
+
+        // flush if the stream is empty
```
vinniefalco, line 1104: LOL the comment is not only unhelpful, but wrong. It should say "buffer" not "stream", and it should say "not empty."

include/boost/json/basic_parser.hpp
```diff
@@ -1094,30 +1100,33 @@ parse_string(
...
-            suspend(state::str3);
+            else
+            {
+                if(BOOST_JSON_UNLIKELY(
+                    ! h.on_string_part(temp, ec_)))
+                    return result::fail;
+            }
+            temp.clear();
         }
+        // if the stream was empty because it was clipped,
```
vinniefalco, line 1122: is empty

test/basic_parser.cpp
```diff
@@ -778,6 +778,41 @@ class basic_parser_test
         BOOST_TEST(! validate(js));
     }
 
+    void
+    testIssue113()
+    {
+        string_view s = 
+            "\"\\r\\n section id='description'>\\r\\nAll        mbers form the uncountable set "
```
vinniefalco, line 785: If this is from the issue can we please use a raw string so we don't have to have all these escapes?

include/boost/json/basic_parser.hpp
sdkrystian: Hah, this proves exactly why this comment is needed. If you look at the `if` statement above, if tests `*cs`, meaning that this statement is only executed when the stream is empty :)

test/basic_parser.cpp
sdkrystian: I used escapes to ensure that the strings are identical (i.e. no extra newlines)

sdkrystian: I used escapes to ensure that the strings are identical (i.e. no extra newlines)

include/boost/json/basic_parser.hpp
sdkrystian: The alternative is the keep `state::str1` and manually call trim at each case label

PR #117, opened 2020-06-29
Treat warnings as errors in CI

sdkrystian:


test/snippets.cpp
```diff
@@ -80,12 +80,10 @@ usingStrings()
         json::string jstr1 = "helloworld";
         json::string jstr2 = "world";
```
vinniefalco, line 81: This would be a better example if the two strings did not share the word "world"

PR #124, opened 2020-07-01
Add storage_ptr constructor to parser

### LINKED ISSUE #79: Add storage_ptr constructor to parser & serializer for internal use

sdkrystian:
Closes #79

### DIFF
--- a/include/boost/json/impl/parser.ipp
+++ b/include/boost/json/impl/parser.ipp
@@ -154,7 +154,14 @@ parser::
 }
 
 parser::
-parser()
+parser() noexcept
+    : parser(storage_ptr())
+{
+}
+
+parser::
+parser(storage_ptr sp) noexcept
+    : rs_(std::move(sp))
 {
     lev_.st = state::need_start;
 }
--- a/include/boost/json/parser.hpp
+++ b/include/boost/json/parser.hpp
@@ -56,14 +56,17 @@ namespace json {
     to bound the amount of work performed in each
     parsing cycle.
 
-    <br>
+    @par Intermediate Storage
 
     The parser may dynamically allocate intermediate
     storage as needed to accommodate the nesting level
-    of the JSON being parsed. This storage is freed
-    when the parser is destroyed, allowing the parser
-    to cheaply re-use this memory when parsing
-    subsequent JSONs, improving performance.
+    of the JSON being parsed. Intermediate storage is
+    allocated using the @ref storage_ptr passed to
+    the constructor; if no such argument is specified,
+    the default memory resource will be used instead.
+    This storage is freed when the parser is destroyed, 
+    allowing the parser to cheaply reuse this memory
+    when parsing subsequent JSONs, improving performance.
 */
 class parser : public basic_parser
 {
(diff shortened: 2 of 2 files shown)


include/boost/json/parser.hpp
```diff
@@ -104,6 +104,21 @@ class parser : public basic_parser
...
+
+        @note
+        Before any JSON can be parsed, the function
+        @ref start must be called.
+
+        @param sp The @ref storage_ptr to use for
+        allocating internal storage.
+    */
+    BOOST_JSON_DECL
+    explicit parser(storage_ptr sp);
```
vinniefalco, line 120: explicit on its own line please

vinniefalco: We should call it "intermediate storage" and make it clear that this is used internally and not for the final DOM. We also need to explain that it is reused between parses, which improves performance.

vinniefalco:
The usage of this constructor needs to be explained in the documentation, along with the implications for performance.

bench/bench.cpp
```diff
@@ -317,11 +317,12 @@ class boost_pool_impl : public any_impl
         string_view s,
         std::size_t repeat) const override
     {
-        parser p;
+        monotonic_resource mr1;
+        parser p(&mr1);
```
vinniefalco, line 321: We shouldn't be doing this , otherwise it won't free the memory when the stack is reallocated. It needs to use the default resource.

include/boost/json/impl/parser.ipp
```diff
@@ -154,7 +154,14 @@ parser::
 }
 
 parser::
-parser()
+parser()
```
vinniefalco, line 157: should this be noexcept?

include/boost/json/impl/parser.ipp
```diff
@@ -154,7 +154,14 @@ parser::
...
 
 parser::
-parser()
+parser() 
+    : parser(storage_ptr())
+{
+}
+
+parser::
+parser(storage_ptr sp)
```
vinniefalco, line 163: should this be noexcept?

include/boost/json/parser.hpp
sdkrystian: Will do

include/boost/json/impl/parser.ipp
sdkrystian: Yup

include/boost/json/impl/parser.ipp
sdkrystian: Yup

sdkrystian:
@vinniefalco In addition to your suggested changes I'll be adding a constructor of this form to `serializer`.

include/boost/json/parser.hpp
```diff
@@ -98,11 +101,39 @@ class parser : public basic_parser
...
+        @note
+        Before any JSON can be parsed, the function
+        @ref start must be called. 
+    */
+    BOOST_JSON_DECL
+    parser() noexcept;
+
+    /** Constructor.
+
+        Constructs a empty parser using the supplied
```
vinniefalco, line 117: Nitpick: I would say specified instead of supplied. But it is too small a distinction to bother changing right now.

include/boost/json/parser.hpp
```diff
@@ -98,11 +101,39 @@ class parser : public basic_parser
...
+        intermediate storage.
+
+        @note
         Before any JSON can be parsed, the function
         @ref start must be called.
+
+        <br>
+
+        The `sp` parameter is only used to
+        allocate intermediate storage; will not be used
```
vinniefalco, line 128: grammar is funky here but we can roll with it.

Issue #115, opened 2020-06-21
hash % buckets() calculation can be sped up


joaquintides:
Your internal hash table uses a prime number table `lists` for bucket array sizes and calculates the position of a value `hash` within the bucket array as `hash % buckets()`. Modulo calculation is an expensive operation involving CPU integer division: turns out this can be _significantly_ sped up by eliminating run-time  modulo ops altogether with the following technique:

* Make `detail::object_impl::table` store the index into the prime number table rather than the prime itself (member `buckets`).
* Replace `hash%buckets()` by a call to a function of the form:
```
std::size_t position(std::size_t hash,std::size_t buckets_size_index)
{
  switch(buckets_size_index){
    default:
    case 0: return 0;
    case 1: return hash%list[1];
    case 2: return hash%list[2];
    case 3: return hash%list[3];
    case 4: return hash%list[4];
    case 5: return hash%list[5];
    case 6: return hash%list[6];
    case 7: return hash%list[7];
    case 8: return hash%list[8];
    case 9: return hash%list[9];
    case 10: return hash%list[10];
    case 11: return hash%list[11];
    case 12: return hash%list[12];
    case 13: return hash%list[13];
    case 14: return hash%list[14];
    case 15: return hash%list[15];
    case 16: return hash%list[16];
    case 17: return hash%list[17];
    case 18: return hash%list[18];
    case 19: return hash%list[19];
    case 20: return hash%list[20];
    case 21: return hash%list[21];
    case 22: return hash%list[22];
    case 23: return hash%list[23];
    case 24: return hash%list[24];
    case 25: return hash%list[25];
    case 26: return hash%list[26];
    case 27: return hash%list[27];
    case 28: return hash%list[28];
    case 29: return hash%list[29];
    case 30: return hash%list[30];
    case 31: return hash%list[31];
    case 32: return hash%list[32];
    case 33: return hash%list[33];
    case 34: return hash%list[34];
    case 35: return hash%list[35];
    case 36: return hash%list[36];
    case 37: return hash%list[37];
    case 38: return hash%list[38];
    case 39: return hash%list[39];
    case 40: return hash%list[40];
    case 41: return hash%list[41];
    case 42: return hash%list[42];
    case 43: return hash%list[43];
    case 44: return hash%list[44];
    case 45: return hash%list[45];
    case 46: return hash%list[46];
    case 47: return hash%list[47];
    case 48: return hash%list[48];
    case 49: return hash%list[49];
[...]

PR #125, opened 2020-07-02
Use constant operand for modulo

sdkrystian:
Implements #115

### DIFF
--- a/include/boost/json/detail/object_impl.hpp
+++ b/include/boost/json/detail/object_impl.hpp
@@ -29,6 +29,7 @@ class object_impl
     using index_t = std::uint32_t;
     static index_t const null_index =
         std::uint32_t(-1);
+    using bucket_size_array = const unsigned long long[67];
 
     BOOST_JSON_DECL
     void
@@ -43,6 +44,7 @@ public:
     inline
     object_impl(
         std::size_t capacity,
+        std::size_t prime_index,
         std::size_t buckets,
         std::uintptr_t salt,
         storage_ptr const& sp);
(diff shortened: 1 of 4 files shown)


vinniefalco:
The commit message body should have the text `fix #115`

include/boost/json/detail/object_impl.hpp
```diff
@@ -133,6 +135,21 @@ class object_impl
         index_t& head,
         value_type& p) noexcept;
 
+    //------------------------------------------------
+    // https://github.com/CPPAlliance/json/issues/115
```
vinniefalco, line 139: dont need this comment

include/boost/json/detail/object_impl.hpp
```diff
@@ -133,6 +135,21 @@ class object_impl
...
+    // 
+    static
+    inline
+    prime_list&
+    primes() noexcept;
+
+    static
+    inline
+    std::size_t
+    position(
```
vinniefalco, line 149: need a better name for this

include/boost/json/detail/object_impl.hpp
```diff
@@ -133,6 +135,21 @@ class object_impl
         index_t& head,
         value_type& p) noexcept;
 
+    //------------------------------------------------
+    // https://github.com/CPPAlliance/json/issues/115
+    // 
+    static
+    inline
+    prime_list&
+    primes() noexcept;
```
vinniefalco, line 144: should this be called `bucket_sizes`?

include/boost/json/detail/object_impl.hpp
sdkrystian: `bucket_index`?

vinniefalco:
These changes look good but we need to explain the benchmark results.

sdkrystian:
@vinniefalco Explain in what way? Like a comment as to why this is better?

PR #116, opened 2020-06-28
Comments, trailing commas, flags, and optimizations

sdkrystian:

PR #126, opened 2020-07-06
Use jump table and better code generation without local_const_stream

sdkrystian:


vinniefalco:
The commit message "Remove local streams from parse null, true, and false" adds no information beyond what is already obvious from the diff. Consider "Better code generation without local_stream" as an alternative.

vinniefalco:
Already in #116

PR #134, opened 2020-08-08
Specialize std::hash for json string

vinniefalco:
The hash function can be optionally salted, for security.

### DIFF
--- a/test/string.cpp
+++ b/test/string.cpp
@@ -2652,6 +2652,9 @@ public:
     void
     testHash()
     {
+        // libstdc++ 4.8 bug
+#if !defined(__GNUC__) || (__GNUC__ > 4 || \
+    (__GNUC__ == 4 && __GNUC_MINOR__ > 8))
         {
             std::unordered_set<string> us;
             us.emplace("first");
@@ -2662,6 +2665,7 @@ public:
                 0,
                 std::hash<string>(32));
         }
+#endif
         {
             std::hash<string> h1(32);
             std::hash<string> h2(h1);


include/boost/json/string.hpp
```diff
@@ -2948,6 +2949,35 @@ operator>(T const& lhs, U const& rhs) noexcept
 } // json
 } // boost
 
+// std::hash specialization
+#ifndef BOOST_JSON_DOCS
+namespace std {
```
sdkrystian, line 2954: This technically isn't needed, we could just do:
```cpp
template<>
struct std::hash<::boost::json::string> { ... };
```

include/boost/json/string.hpp
```diff
@@ -2948,6 +2949,35 @@ operator>(T const& lhs, U const& rhs) noexcept
...
+template<>
+struct hash<::boost::json::string>
+{
+    hash() = default;
+    hash(hash const&) = default;
+    hash& operator=(hash const&) = default;
+
+    explicit
+    hash(std::size_t salt)
```
sdkrystian, line 2963: This must be marked `noexcept` (see http://eel.is/c++draft/unord.hash#3)

include/boost/json/detail/impl/digest.ipp
```diff
@@ -0,0 +1,66 @@
...
+
+inline
+std::uint64_t
+digest_impl(
+    char const* s,
+    std::size_t n,
+    std::size_t salt,
+    std::true_type) noexcept
+{
+    std::uint64_t const prime = 0x100000001B3ULL;
```
sdkrystian, line 41: We could just use `BOOST_JSON_ARCH` to set the constants instead of having two overloads

sdkrystian: E.g. 
```cpp
#if BOOST_JSON_ARCH == 64
std::uint64_t const prime = 0x100000001B3ULL;
std::uint64_t hash  = 0xcbf29ce484222325ULL;
#else 
std::uint32_t const prime = 0x01000193UL;
std::uint32_t hash  = 0x811C9DC5UL;
#endif 
...
```

test/string.cpp
```diff
@@ -2648,6 +2649,27 @@ class string_test
         BOOST_TEST(ss.str() == s);
     }
 
+    void
+    testHash()
+    {
```
sdkrystian, line 2654: These unused variables are causing test failures (`-Wextra`)

PR #136, opened 2020-08-08
Bug fixes & coverage improvements

sdkrystian:


vinniefalco:
I would like to see all the constructors for `parse_options` removed, and the tests fixed.

vinniefalco:
Merged most of it (not the `parse_options` bit), thanks.

PR #137, opened 2020-08-08
Parser optimizations

sdkrystian:


sdkrystian:
This PR was rebased on top of #136, so really only the 5 most recent commits should be merged

include/boost/json/detail/basic_parser.hpp
```diff
@@ -108,59 +100,110 @@ class basic_parser
...
+
+    BOOST_NOINLINE
+    inline
+    std::nullptr_t
+    partial_if_more(
+        const char* p, 
+        state st);
+
+    BOOST_NOINLINE
+    inline
```
vinniefalco, line 135: why `BOOST_NOINLINE` followed by `inline` ?

include/boost/json/detail/basic_parser.hpp
```diff
@@ -199,7 +198,8 @@ class basic_parser
     template<bool StackEmpty, class Handler>
     const char* parse_false(Handler& h, const char* p);
     
-    template<bool StackEmpty, bool AllowInvalid, class Handler>
+    template<bool StackEmpty, bool IsKey,
+        bool AllowInvalid, class Handler>
```
vinniefalco, line 202: I think we should rename `AllowInvalid` to `AllowBadUTF8`. This change should be in its own commit.

vinniefalco:
I think the "Better key parsing" commit should be removed entirely. It is an optimization, and I do not want to merge any more optimizations, there have been enough already. Any new optimizations should be submitted only when the following conditions have all been met:
1. Boost.JSON is accepted in review,
2. There have been no logic or feature changes merged for at least 60 days, and
3. The change is accompanied by comprehensive benchmarks in a stable environment which demonstrate the benefit.
4. A measurement is provided of the change in the size of the emitted binary with optimizations turned on.

include/boost/json/detail/basic_parser.hpp
sdkrystian: We want the ODR benefits of the `inline` keyword (multiple definitions), but we don't actually want the function to be inlined.

vinniefalco:
Too many changes and too many unnecessary changes, sorry

PR #139, opened 2020-08-10
Remove constructor from parse_options and add docs

sdkrystian:


test/basic_parser.cpp
```diff
@@ -116,6 +116,19 @@ validate( string_view s )
...
+    bool comments,
+    bool commas,
+    bool utf8)
+{
+    parse_options opt;
+    opt.allow_comments = comments;
+    opt.allow_trailing_commas = commas;
+    opt.allow_invalid_utf8 = utf8;
+    return opt;
+}
```
glenfe, line 130: Why do you need this?  `parse_options{ false, true, false }` works with `parse_options` as you've currently defined it.

sdkrystian: @glenfe Not for C++11; `parse_options` has default member initializers and thus is not an aggregate. C++14 and above doesn't have this problem.

vinniefalco:
Approved

PR #141, opened 2020-08-10
Refactor stream aliasing, suspend, and error paths

sdkrystian:


sdkrystian:
A few changes that should be made:
- Pass/return `const_stream` instead of `const char*`. This will require an extra `mov` for calls, but it eliminates a `mov` from memory (likely cache, but still slower) each time a function is called.
-  Make `end_` `const` and don't utilize it in assignment/copy construction
- Set `p_` to `end_` to indicate failure, and write the point at which parsing stopped to a data member.

include/boost/json/basic_parser.hpp
```diff
@@ -218,47 +218,106 @@ reserve()
...
+    return nullptr;
 }
 
-// return `false` if fully consumed
-bool
+std::nullptr_t
+basic_parser::
+fail(
+    const char* p, 
+    error err) noexcept
```
vinniefalco, line 246: `error ev`

include/boost/json/basic_parser.hpp
```diff
@@ -296,103 +354,86 @@ parse_comment(
...
     {
     case '/':
         ++cs;
 do_com2:
         remain = cs.remain();
-        end_tok = remain ? static_cast<const char*>(
-            std::memchr(cs.data(), '\n', remain)) : nullptr;
-        if(BOOST_JSON_UNLIKELY(! end_tok))
+        cs = remain ? static_cast<const char*>(
+            std::memchr(cs.begin(), '\n', remain)) : nullptr;
```
vinniefalco, line 365:
```
if( remain > 0 )
  cs = static_cast<const char*>(
    std::memchr(cs.data(), '\n', remain) );
else
  cs = nullptr;
```

sdkrystian: Yeah, I'll be moving this out of the function.

include/boost/json/detail/basic_parser.hpp
```diff
@@ -138,63 +129,109 @@ class basic_parser
...
-    inline void suspend(state st, number const& num);
-    inline bool skip_white(const_stream& cs);
-
-    result syntax_error(const_stream&);
+    inline const char* canary();
+
+    BOOST_NOINLINE
+    inline
+    const char*
+    propagate(state st);
```
vinniefalco, line 144: This is not descriptive enough

include/boost/json/basic_parser.hpp
```diff
@@ -272,105 +359,92 @@ parse_comment(
...
     case '/':
         ++cs;
 do_com2:
+        // KRYSTIAN TODO: this is a mess, we have to fix this
         remain = cs.remain();
-        end_tok = remain ? static_cast<const char*>(
-            std::memchr(cs.data(), '\n', remain)) : nullptr;
-        if(BOOST_JSON_UNLIKELY(! end_tok))
+        cs = remain ? static_cast<const char*>(
+            std::memchr(cs.begin(), '\n', remain)) : canary();
```
vinniefalco, line 371: no ternary please

sdkrystian: Going to address this in another commit, later on

include/boost/json/detail/basic_parser.hpp
sdkrystian: IMO it's the least misleading. I'll add docs saying what it actually does.

Issue #101, opened 2020-05-28
Support raw number literals in basic_parser


rtobar:
The current set of virtual functions that subclasses of `basic_parser` need to implement provides users with entry points for receiving `int64`, `uint64` and `double` numerical values out of the parsing process.

JSON content however might contain numerical values outside of these types' ranges, or that cannot be exactly represented by any of them (e.g., `"1e400"`, `"1e-400"` or `"3.141592653589793238462643383279"`). In these cases users loose have no access to the original values as seen in the JSON stream, and can only access the values as calculated by `basic_parser`, which will already be overflown, underflown, or otherwise misrepresented, therefore loosing information (e.g., `Infinity`, `0` and `3.141592653589793115997963468544185161590576171875` for the examples above). Adding a way for users to get access to the original string values for those numbers would allow them to access the original values without information loss.

Just for reference, I only have experience with yajl, which gives user the option to provide callbacks either for integer/double values, or for numbers-as-string. The two modes are mutually exclusive, so when using the latter users have to do their own number parsing. I have no idea how this support would be best implemented in this library though, which otherwise has given me very good results so far.

vinniefalco:
This can be done by adding a handler member `on_number_part` for partial strings, and then adding a `string_view` to the existing on_number handlers. The compiler should in theory optimize it away if it is not used, since the handler's definition is visible to the parsing function.

vinniefalco:
@sdkrystian I believe this was merged?

sdkrystian:
@vinniefalco Oh, oops. I have the branch, but it never got merged. I'll rebase it and open a PR.

vinniefalco:
This is merged

PR #142, opened 2020-08-10
Pass raw number literals to handlers

sdkrystian:
Needs tests

### DIFF
--- a/include/boost/json/basic_parser.hpp
+++ b/include/boost/json/basic_parser.hpp
@@ -2321,6 +2321,7 @@ parse_number(
     constexpr bool nonzero_first = First == '+';
     number num;
     detail::local_const_stream cs(cs0);
+    const char* begin = cs.data();
     if(StackEmpty || st_.empty())
     {
         num.bias = 0;
@@ -2490,6 +2491,10 @@ do_num1:
     }
     else
     {
+        if(BOOST_JSON_UNLIKELY(
+            ! h.on_number_part(
+                {begin, cs.used(begin)}, ec_)))
+            return result::fail;
         if(more_)
             suspend(state::num1, num);
         return result::partial;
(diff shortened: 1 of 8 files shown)


include/boost/json/basic_parser.hpp
```diff
@@ -2490,6 +2491,10 @@ parse_number(
     }
     else
     {
+        if(BOOST_JSON_UNLIKELY(
+            ! h.on_number_part(
+                {begin, cs.used(begin)}, ec_)))
```
vinniefalco, line 2496: @glenfe @pdimov As written, the init-list allows `{begin, cs.used(begin)}` allows `on_number_part` to be declared with an argument type that may not be compatible with string view. Is this desirable? Or should the code above be written as:
```
            ! h.on_number_part(
                string_view(begin, cs.used(begin)),
                ec_)))
```

pdimov: Hard to say. If we require `on_number_part` to take exactly a `string_view` parameter, the above is fine. If we always pass a `string_view`, this allows `on_number_part` to take something convertible from `string_view`, which is probably also undesirable.

sdkrystian:
Addresses #101

include/boost/json/basic_parser.hpp
vinniefalco: I haven't bothered to write code to force the handler to have particular signatures (or else get a compile errror) because it doesn't seem worth it.

sdkrystian: IMO it's worth it because by restricting the signature we can potentially optimize around it.

vinniefalco: we can restrict the signature without writing any code, just say it in the docs "the signature must match"

vinniefalco:
Approved

PR #144, opened 2020-08-10
Tuple-like doc work

### LINKED ISSUE #120: Docs for tuple-like interface
`get` will be missing some javadocs when the PR is merged. It should probably be done like this:
```
#ifdef BOOST_JSON_DOCS

/** Tuple-like interface
*/
template< std::size_t I, class T >
__see_below__
get( T&& );

#else

// actual defintions

#endif
```

sdkrystian:
Fixes #120

### DIFF
--- a/include/boost/json/value.hpp
+++ b/include/boost/json/value.hpp
@@ -2839,7 +2839,8 @@ get(key_value_pair const&) noexcept ->
         string_view const,
         value const&>::type
 {
-    BOOST_STATIC_ASSERT(I == -1u);
+    static_assert(I == 0, 
+        "key_value_pair index out of range");
 }
 
 template<std::size_t I>
@@ -2849,7 +2850,8 @@ get(key_value_pair&) noexcept ->
         string_view const,
         value&>::type
 {
-    BOOST_STATIC_ASSERT(I == -1u);
+    static_assert(I == 0, 
+        "key_value_pair index out of range");
 }
 
 template<std::size_t I>
@@ -2859,7 +2861,8 @@ get(key_value_pair&&) noexcept ->
         string_view const,
         value&&>::type
 {
-    BOOST_STATIC_ASSERT(I == -1u);
+    static_assert(I == 0, 
+        "key_value_pair index out of range");
 }
 
 /** Extracts a key_value_pair's key using tuple-like interface
--- a/include/boost/json/value.hpp
+++ b/include/boost/json/value.hpp
@@ -1,5 +1,6 @@
 //
 // Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+// Copyright (c) 2020 Krystian Stasiowski (sdkrystian@gmail.com)
 //
 // Distributed under the Boost Software License, Version 1.0. (See accompanying
 // file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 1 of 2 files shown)


include/boost/json/value.hpp
```diff
@@ -2832,6 +2832,65 @@ struct key_value_pair
...
+
+    @return `kvp.key()` if `I` is 0, or `kvp.value()` 
+    if `I` is `1`.
+
+    @param kvp The `key_value_pair` object
+    to access.
+*/
+template<
+    class T,
+    std::size_t I>
```
vinniefalco, line 2888: Shouldn't the `I` come first?

include/boost/json/value.hpp
```diff
@@ -2832,6 +2832,65 @@ struct key_value_pair
...
+    the argument is a non-const rvalue, and 
+    `const value&` otherwise..
+    Any other value for `I` is ill-formed.
+
+    @tparam I The element index to access.
+
+    @par Constraints
+
+    `T` is (possibly cv- or reference-qualified)
+    `key_value_pair`.
```
vinniefalco, line 2878: Use `@ref key_value_pair` and it will link

include/boost/json/value.hpp
```diff
@@ -2832,6 +2832,65 @@ struct key_value_pair
...
+
+    @return `kvp.key()` if `I` is 0, or `kvp.value()` 
+    if `I` is `1`.
+
+    @param kvp The `key_value_pair` object
```
vinniefalco, line 2883: Use `@ref key_value_pair`

include/boost/json/value.hpp
```diff
@@ -2832,6 +2832,65 @@ struct key_value_pair
 
 //----------------------------------------------------------
 
+#ifdef BOOST_JSON_DOCS
+
+/** Tuple-like element access.
+    
+    This overload permits the key and value
+    of a `key_value_pair` to be accessed
```
vinniefalco, line 2840:
```
This function returns members of a `@ref key_value_pair` by index.
```

include/boost/json/value.hpp
```diff
@@ -2832,6 +2832,65 @@ struct key_value_pair
...
+    by index. For example:
+
+    @code 
+
+    key_value_pair kvp("num", 42);
+    
+    auto key = get<0>(kvp);
+    auto value = get<1>(kvp);
```
vinniefalco, line 2848: We shouldn't use auto in docs, as there is no benefit to doing so and because showing the actual type is more informative for the reader:
```
string_view const key = get<0>(kvp);
value& jv = get<1>(kvp);

File: include/boost/json/value.hpp [hunk 6]
```diff
@@ -2832,6 +2832,65 @@ struct key_value_pair
...
+
+    @endcode
+
+    In addition to calling the function directly,
+    this function allows `key_value_pair` objects
+    to initialize structured bindings:
```
vinniefalco, line 2854:
```
@par Structured Bindings

When using C++17 or greater, objects of type @ref key_value_pair
may be used to initialize structured bindings:
```

include/boost/json/value.hpp
```diff
@@ -2832,6 +2832,65 @@ struct key_value_pair
...
+    @endcode
+
+    The return type of this template
+    depends upon the value of `I`.
+    If `I` is 0, the return type is `const string_view`.
+    If `I` is 1, the return type is `value&` if
+    the argument is a non-const lvalue, `value&&` if 
+    the argument is a non-const rvalue, and 
+    `const value&` otherwise..
+    Any other value for `I` is ill-formed.
```
vinniefalco, line 2871:
```
Depending on the value of `I`, the return type will be:

@li `string_view const` if `I == 0`,

@li `value&` or `value const&` if `I == 1`, otherwise

@li the program is ill-formed.
```

Note that the docca toolchain has a preference for east-const, so we need to be consistent in user-facing program code and identifiers.

include/boost/json/value.hpp
vinniefalco: Actually the convention in the docs (I think) is to use a C++ expression here. So we might write something like:
```
@par Constraints

`std::is_same< std::remove_cvref_t<T>, key_value_pair >`
```

include/boost/json/value.hpp
```diff
@@ -2832,6 +2832,65 @@ struct key_value_pair
 
 //----------------------------------------------------------
 
+#ifdef BOOST_JSON_DOCS
```
vinniefalco, line 2835: When we use `BOOST_JSON_DOCS` we need to strive to make the conditionally compiled section as small as possible. This exposes more of the doc material to the compiler and helps catch errors. For example we would prefer:
```
#ifdef BOOST_JSON_DOCS
__see_below__
#else
ugly_type
#endif
great_function( complicated_args... );
```

over

```
#ifdef BOOST_JSON_DOCS
__see_below__
great_function( complicated_args... );
#else
ugly_type
great_function( complicated_args... );
#endif
```

So I think, we should try to apply this to the code here (not sure what the best way is).

include/boost/json/value.hpp
```diff
@@ -2832,6 +2832,65 @@ struct key_value_pair
...
+
+    In addition to calling the function directly,
+    this function allows `key_value_pair` objects
+    to initialize structured bindings:
+
+    @code
+
+    key_value_pair kvp("num", 42);
+
+    auto& [key, value] = kvp;
```
vinniefalco, line 2860: Keep this as `auto` of course

include/boost/json/value.hpp
sdkrystian: Oh, oops

include/boost/json/value.hpp
sdkrystian: Eh... I'd rather say "key and value". Members is too broad, and there's only two that we provide access to.

vinniefalco: OK

vinniefalco: The reason I worded it that way is to avoid repetition of the words "key" and "value" but we can go with your wording too.

vinniefalco:
Approved

PR #147, opened 2020-08-10
Fix integer benchmark ranges

sdkrystian:
`integers-32` contains numbers in the range `-2^32 to 2^32 - 1` and `integers-64` contains numbers in the range `-2^64 to 2^64 - 1` that don't fall within the range of `integers-32`. Numbers are generated randomly based on length, so it's a good mix of short and long numbers.


sdkrystian:
Updated to include `-2^32 to 2^32 - 1` in `integers-64`

sdkrystian:
Whoops, accidently deleted the branch when pruning stale branches.

vinniefalco:
Is this relevant now that we are using an updated set of benchmark files?

sdkrystian:
Nope, this can be closed.

PR #148, opened 2020-08-11
Use the last object key

### LINKED ISSUE #71: take the last duplicate key
_"...most popular implementations (including the ECMAScript specification which is implemented in modern browsers) follow the rule of taking only the last key-value pair..."_

sdkrystian:
Fixes #71

### DIFF
--- a/include/boost/json/detail/impl/object_impl.ipp
+++ b/include/boost/json/detail/impl/object_impl.ipp
@@ -70,36 +70,30 @@ void
 object_impl::
 build() noexcept
 {
-    auto end = this->end();
-    for(auto p = begin(); p != end;)
+    // must work when table pointer is null
+    auto const first = begin();
+    for(auto last = end(); last > first;)
     {
+        --last;
+        auto head = &bucket(last->key());
+        auto i = *head;
+        while(i != null_index &&
+            get(i).key() != last->key())
+            i = next(get(i));
+        if(i != null_index)
         {
-            auto& head = bucket(p->key());
-            auto i = head;
-            while(i != null_index &&
-                get(i).key() != p->key())
-                i = next(get(i));
-            if(i == null_index)
-            {
-                next(*p) = head;
-                head = index_of(*p);
-                ++p;
-                continue;
-            }
-        }
-        p->~value_type();
-        --tab_->size;
-        --end;
-        if(p != end)
-        {
+            // handle duplicate
+            last->~value_type();
             std::memcpy(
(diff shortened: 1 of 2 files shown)


include/boost/json/detail/impl/object_impl.ipp
```diff
@@ -70,36 +70,29 @@ void
 object_impl::
 build() noexcept
 {
-    auto end = this->end();
-    for(auto p = begin(); p != end;)
+    const auto first = begin() - 1;
```
vinniefalco, line 73: Do we need to `BOOST_ASSERT` that the list is not empty?

vinniefalco:
Approved

include/boost/json/detail/impl/object_impl.ipp
sdkrystian: No, `first < last` is `false` if that happens.

include/boost/json/detail/impl/object_impl.ipp
```diff
@@ -70,36 +70,31 @@ void
 object_impl::
 build() noexcept
 {
-    auto end = this->end();
-    for(auto p = begin(); p != end;)
+    if(! size())
+        return;
+    const auto first = begin() - 1;
+    auto last = end() - 1;
+    for(; last > first; --last)
```
vinniefalco, line 77: Can we avoid the extra conditional with this
```
auto const first = begin();
auto last = end();
for(; last-- > first;)
{
    ...
```
?

sdkrystian: No, we still need to do an iteration when `last == first`

vinniefalco: Okay that was no good, this should work though:
```
7
// must work when table pointer is null
auto const first = begin();
for( auto last = end(); last > first; )
{
    --last;
    ...
```

sdkrystian: Did you change this on your branch or should I change it here?

vinniefalco: I prefer not to make these kinds of changes on my own, because then it can be confusing if you are also making the change, but I already merged it, and maybe you found a bug so now you have a branch that is different than mine. Since you're the lead on this pull request, it is better if you are the only one applying changes to it.

sdkrystian: Copy that

PR #159, opened 2020-08-16
Fix unintended move from string initializer list elements

### LINKED ISSUE #158: Surprising move-from behavior when initializing a json::object
Thank you for this well designed, well documented, and easy to use library! It is quite pleasant to use. 

I wonder if this behavior is as you designed? The documentation of 'Initializer Lists' indicates that "Elements that are rvalues will be moved upon initialization," but this program exhibits behavior that is surprising to me:

```c++
#include <boost/json.hpp>
#include <iostream>

int
main()
{
  using namespace boost::json;

  string s1{ "hello world" };

  // obj1 will move from s1, invalidating s1
  object obj1{ { { "s1", s1 } } };

  if (s1 == "hello world")
    std::cout << "OK\n";
  else
    std::cout << "FAIL\n";

  //

  // making s2 const inhibits the move-from
  string const s2{ "hello world" };

  object obj2{ { { "s2", s2 } } };

  if (s2 == "hello world")
    std::cout << "OK\n";
  else
    std::cout << "FAIL\n";

  //
  // program prints:
  //
  // FAIL
  // OK
  //
}
```

Unless I'm mistaken, `s1` is not an rvalue, and should not be moved from. Making the variable `const` inhibits the move.

I hope this is not intended behavior, because it does make it far too easy to shoot myself in the foot with use-after-move issues, especially since no `std::move` is visible in my own code. Thank you again!

sdkrystian:
Fixes #158

### DIFF
--- a/test/value_ref.cpp
+++ b/test/value_ref.cpp
@@ -324,6 +324,12 @@ public:
             BOOST_TEST(a.empty());
         }
         {
+            string a = "abcdefghijklmnopqrstuvwxyz";
+            BOOST_TEST(!a.empty());
+            array b{a, string()};
+            BOOST_TEST(!a.empty());
+        }
+        {
             array a{value()};
             BOOST_TEST(a.data() != nullptr);
             array b{std::move(a), array()};
--- a/include/boost/json/value_ref.hpp
+++ b/include/boost/json/value_ref.hpp
@@ -155,16 +155,14 @@ public:
         ,typename std::enable_if<
             (! std::is_constructible<
                 string_view, T>::value ||
-            std::is_same<string, 
-                detail::remove_cvref<T>>::value) &&
+            std::is_same<string, T>::value) &&
             ! std::is_same<bool,
                 detail::remove_cvref<T>>::value &&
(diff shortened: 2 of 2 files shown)


test/value_ref.cpp
```diff
@@ -323,6 +323,12 @@ class value_ref_test
             array b{std::move(a), string()};
             BOOST_TEST(a.empty());
         }
+        {
+            string a = "abcdefghijklmnopqrstuvwxyz";
+            BOOST_TEST(!a.empty());
+            array b{a, string()};
```
vinniefalco, line 329: I hate this form of initialization. I prefer
```
array b = { a, string() };
```
or
```
array b( { a, string() } );
```

sdkrystian: Meh, it's the style we use in all the other tests. If you want, I can change them all

vinniefalco: No I'm just grumbling. The user reported a const string though, should we test that too?

vinniefalco: I added a test for `string const`

PR #160, opened 2020-08-16
Add value_builder

vinniefalco:
This makes the internal parser handler a public type.

### DIFF
--- a/include/boost/json/value_builder.hpp
+++ b/include/boost/json/value_builder.hpp
@@ -1,0 +1,555 @@
+//
+// Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/cppalliance/json
+//
+
+#ifndef BOOST_JSON_VALUE_BUILDER_HPP
+#define BOOST_JSON_VALUE_BUILDER_HPP
+
+#include <boost/json/detail/config.hpp>
+#include <boost/json/error.hpp>
+#include <boost/json/storage_ptr.hpp>
+#include <boost/json/value.hpp>
+#include <boost/json/detail/raw_stack.hpp>
+#include <stddef.h>
+
+namespace boost {
+namespace json {
+
+//----------------------------------------------------------
+
+/** A factory for building a value DOM.
+
+    A value builder implements an algorithm for
+    efficiently constructing a @ref value from an
+    external source (provided by the caller).
+
+    The builder uses a dynamically allocated internal
+    storage to hold portions of the document, allowing
+    complete objects and arrays to be constructed using
+    a single allocation when their contents are
+    eventually known. This internal storage is reused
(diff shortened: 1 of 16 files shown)


vinniefalco:
Testing

include/boost/json/value_builder.hpp
```diff
@@ -0,0 +1,556 @@
...
+    level lev_;
+
+public:
+    /** Destructor.
+
+        All dynamically allocated memory, including
+        any partially built results, is freed.
+    */
+    BOOST_JSON_DECL
+    virtual
```
sdkrystian, line 73: We moved away from the inheritance model, should this still be virtual?

include/boost/json/value_builder.hpp
```diff
@@ -0,0 +1,556 @@
...
+        returns `true`.
+
+        @return `true` on success.
+
+        @param s The characters to append. This may be empty.
+
+        @param ec Set to the error, if any occurred.
+    */
+    bool
+    on_number_part(
```
sdkrystian, line 357: Instead of manually casting to `void`, we can just leave the parameters unnamed.

include/boost/json/value_builder.hpp
```diff
@@ -0,0 +1,556 @@
...
+        This function builds a number from the specified
+        value and adds it as the top-level element of the
+        resulting value, or as the next element of the
+        current object or array being built.
+
+        @return `true` on success.
+
+        @param i The unsigned integer to build.
+
+        @param s The characters to append. This value is ignored.
```
sdkrystian, line 399: There is no parameter named `s` here.

include/boost/json/value_builder.hpp
```diff
@@ -0,0 +1,556 @@
...
+
+        @param ec Set to the error, if any occurred.
+    */
+    BOOST_JSON_DECL
+    bool
+    on_double(
+        double d,
+        string_view,
```
sdkrystian, line 429: There is no parameter named `s` here.

include/boost/json/value_builder.hpp
```diff
@@ -0,0 +1,556 @@
...
+    */
+    BOOST_JSON_DECL
+    bool
+    on_bool(
+        bool b,
+        error_code& ec);
+
+    /** Build a null.
+
+        This function builds a nullfrom the specified
```
sdkrystian, line 453: Missing space in "nullfrom", and there is no parameter name `ec` here.

include/boost/json/value_builder.hpp
```diff
@@ -0,0 +1,556 @@
...
+
+        This function has no effect and always returns `true`.
+
+        @param s The characters to append. This value is ignored.
+
+        @param ec Set to the error, if any occurred.
+    */
+    bool
+    on_comment_part(
+        string_view s,
```
sdkrystian, line 476: We can leave the parameters unnamed, as with all the other unused parameters.

include/boost/json/value_builder.hpp
vinniefalco: Then the docs won't be right

include/boost/json/value_builder.hpp
vinniefalco: Nope

PR #168, opened 2020-08-18
fix #166: do not pass a nullptr to deallocate

### LINKED ISSUE #166: null pointer illegally passed to deallocate
The following code (straight from the documentation):

```cpp
error_code ec;
value jv = parse( "[1,2,3,4,5]", ec );
```
compiled with undefined behaviour sanitizer, causes the following output:

> ../include/boost/json/detail/impl/stack.ipp:38:21: runtime error: null pointer passed as argument 1, which is declared to never be null
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/memory_resource:103:20: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../include/boost/json/detail/impl/stack.ipp:38:21 in 

It looks like https://github.com/CPPAlliance/json/blob/400136a292fda23452c5d914cda7fef70939a2b5/include/boost/json/detail/impl/stack.ipp#L38 uses buf_ without checking it for null.

I also wonder if there is perhaps a copy paste error here, using size_ instead of cap_? Should it perhaps be
```diff
--- a/include/boost/json/detail/impl/stack.ipp
+++ b/include/boost/json/detail/impl/stack.ipp
@@ -35,7 +35,8 @@ reserve(std::size_t n)
     if(buf_ && size_ > 0)
         std::memcpy(
             buf, buf_, size_);
-    sp_->deallocate(buf_, size_);
+    if(buf_)
+        sp_->deallocate(buf_, cap_);
     buf_ = buf;
     cap_ = n;
 }

```

pauldreik:
Closes #166

### DIFF
--- a/include/boost/json/detail/impl/stack.ipp
+++ b/include/boost/json/detail/impl/stack.ipp
@@ -35,7 +35,8 @@ reserve(std::size_t n)
     if(buf_ && size_ > 0)
         std::memcpy(
             buf, buf_, size_);
-    sp_->deallocate(buf_, size_);
+    if(buf_)
+        sp_->deallocate(buf_, cap_);
     buf_ = buf;
     cap_ = n;
 }
--- a/include/boost/json/detail/impl/stack.ipp
+++ b/include/boost/json/detail/impl/stack.ipp
@@ -32,11 +32,12 @@ reserve(std::size_t n)
         return;
     auto const buf = static_cast<
         char*>(sp_->allocate(n));
-    if(buf_ && size_ > 0)
-        std::memcpy(
-            buf, buf_, size_);
     if(buf_)
+    {
+        if(size_ > 0)
+            std::memcpy(buf, buf_, size_);
         sp_->deallocate(buf_, cap_);
+    }
     buf_ = buf;
     cap_ = n;
 }


include/boost/json/detail/impl/stack.ipp
```diff
@@ -35,7 +35,8 @@ reserve(std::size_t n)
     if(buf_ && size_ > 0)
         std::memcpy(
             buf, buf_, size_);
-    sp_->deallocate(buf_, size_);
+    if(buf_)
+        sp_->deallocate(buf_, cap_);
```
vinniefalco, line 39: Consider folding the condition?
```
    if( buf_ )
    {
        if( size_ > 0 )
            std::memcpy( buf, buf_, size_ );
        sp_->deallocate(buf_, size_);
    }
```

pauldreik: Or even skip the check for buf_, replacing it with the size check and an assert on buf_?

pauldreik: The idea being that if size is >0, buf_ should definitely be nonnull or else something has gone haywire somewhere else.

vinniefalco: That may be possible, but you would have to look at the entire class to know if that precondition is correct. If you do that, use `BOOST_ASSERT`. However, just a guess but it might not be possible, as `size_` can be zero while `buf_` is nonzero. This is the case when the stack grows, and then everything is popped off, leaving it with a non-zero capacity but a zero size.

pauldreik: I will go for the folded condition :-)

pauldreik:
A thought on the size check. I guess it is there to avoid the expense of unnecessarily calling the memcpy function. But how common is that case with zero size and nonzero buffer? The check introduces a branch, which might be bad for performance.

vinniefalco:
> how common is that case with zero size and nonzero buffer

Very common actually. It happens every time a parser is re-used after retrieving a value. I'm okay with removing the check for `size_` though, as `memcpy` is safe to call with a zero size. Up to you.

pauldreik:
I think the check should remain in that case. Either way, I wouldn't
remove it without benchmarking first. So please go ahead and merge this
fix if you are happy with it.

vinniefalco:
I think it is good thanks

PR #178, opened 2020-08-20
Check for overflow in basic_parser

sdkrystian:


include/boost/json/basic_parser.hpp
```diff
@@ -197,10 +197,16 @@ reserve()
...
+        // document parsing state
+        sizeof(state) + 
+        // array and object state + size
+        (sizeof(state) + sizeof(std::size_t)) * depth() +
+        // value parsing state
+        sizeof(state) + 
+        // string size
+        sizeof(std::size_t) +
+        // comment/utf8 state
+        sizeof(state));
```
vinniefalco, line 209: this is harder to read please put the comments back on the same line. reformat line 203 thusly:
```
(sizeof(state) +
    sizeof(std::size_t)) * depth() +  // array and object state + size
```

PR #188, opened 2020-08-23
Doc work

sdkrystian:


vinniefalco:
I don't think we can remove "Using Numbers." We have to fill it in, if for no other reason than to explain the range and distinction for how the container treats numbers as different variant discriminators.

vinniefalco:
If you have changes please put them in a new PR since I am already working on merging this.

doc/qbk/02_10_init_lists.qbk
```diff
@@ -12,7 +12,7 @@
 [section:initializer_lists Initializer Lists]
 
 Initializer lists can be used to construct or
-assign objects of type __value__:
+assign JSON __value__'s:
```
vinniefalco, line 15: no this doesn't work, `values's` should not be possessive.

PR #200, opened 2020-08-24
Doc work

### LINKED ISSUE #195: Clarifying Memory Resource Base Class in Dox
On this page: http://vinniefalco.github.io/doc/json/json/usage/allocators.html :

memory_resource is listed in Table 2. Functions and Types and is also mentioned in the first paragraph of _Custom Memory Resources_.

I think it would be clearer in both cases to refer to it as std::pmr::memory_resource, because it isn't clear on casual reading that this is type defined by the standard rather than by the library and I feel that this important.

sdkrystian:
close #195


vinniefalco:
What about all the other occurrences of BeastLounge?

PR #201, opened 2020-08-24
Parser interface

sdkrystian:


sdkrystian:
@pauldreik Do we need `parser::release` to throw to not break fuzzing?

include/boost/json/impl/parse.ipp
```diff
@@ -0,0 +1,55 @@
...
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/cppalliance/json
+//
+
+#ifndef BOOST_JSON_IMPL_PARSE_IPP
+#define BOOST_JSON_IMPL_PARSE_IPP
+
+#include <boost/json/parse.hpp>
```
vinniefalco, line 13: You don't need to include this because .ipp files are assumed to come after the include of the corresponding public .hpp

include/boost/json/impl/parse.ipp
```diff
@@ -0,0 +1,55 @@
+//
+// Copyright (c) 2020 Krystian Stasiowski (sdkrystian@gmail.com)
```
vinniefalco, line 2: hey now... I wrote everything in this file

include/boost/json/parse.hpp
```diff
@@ -0,0 +1,98 @@
+//
+// Copyright (c) 2020 Krystian Stasiowski (sdkrystian@gmail.com)
```
vinniefalco, line 2: hey now, I wrote everything in this file. moving it to a new file doesn't mean you have the sole authorship

include/boost/json/parse.hpp
```diff
@@ -0,0 +1,98 @@
...
+// Official repository: https://github.com/cppalliance/json
+//
+
+#ifndef BOOST_JSON_PARSE_HPP
+#define BOOST_JSON_PARSE_HPP
+
+#include <boost/json/detail/config.hpp>
+#include <boost/json/error.hpp>
+#include <boost/json/storage_ptr.hpp>
+#include <boost/json/value.hpp>
```
vinniefalco, line 16: might as well add `<boost/json/string_view.hpp>` while we're at it

include/boost/json/parse.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+namespace boost {
+namespace json {
+
+/** Parse a string of JSON.
+
+    This function parses an entire single string in
```
vinniefalco, line 23: remove the word "single"

include/boost/json/parse.hpp
```diff
@@ -0,0 +1,98 @@
...
+BOOST_JSON_DECL
+value
+parse(
+    string_view s,
+    error_code& ec,
+    storage_ptr sp = {});
+
+/** Parse a string of JSON.
+
+    This function parses an entire single string in
```
vinniefalco, line 59: remove the word "single"

include/boost/json/parse.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+    @param s The string to parse.
+
+    @param sp The memory resource that the new value and all of
+    its elements will use. If this parameter is omitted,
+    the default memory resource is used.
+
+    @throw system_error Thrown on failure.
+
+    @return A value representing the parsed JSON upon success.
```
vinniefalco, line 83: Should the return value annotation come before the params?

include/boost/json/parse.hpp
```diff
@@ -0,0 +1,98 @@
...
+    storage_ptr sp = {});
+
+} // json
+} // boost
+
+#ifdef BOOST_JSON_HEADER_ONLY
+#include <boost/json/impl/parse.ipp>
+#endif
+
+#endif
```
vinniefalco, line 98: you always do this..

include/boost/json/parse.hpp
sdkrystian: omg sorry I'll change that

include/boost/json/parse.hpp
vinniefalco: (the line is missing a carriage return)

include/boost/json/parse.hpp
sdkrystian: Oops, yup

include/boost/json/parse.hpp
sdkrystian: LOL my bad

fuzzing/fuzz_parser.cpp
```diff
@@ -29,7 +29,7 @@ fuzz_parser(string_view sv)
     // Take ownership of the resulting value.
     if(! ec)
     {
-        value jv = p.release();
+        value jv = p.release( ec );
         return jv.is_number();
```
vinniefalco, line 33: you have to check `ec` here. correct code:
```
    // Take ownership of the resulting value.
    if(! ec)
    {
        value jv = p.release( ec );
        if(! ec )
            return jv.is_number();
    }
    return false;
```

I fixed it in my copy

pauldreik:
> @pauldreik Do we need `parser::release` to throw to not break fuzzing?

No, it should not matter for correctness. But for fuzzer performance it is better to avoid throwing exceptions. Fuzzing causes problematic input all the time, so the exceptional case is the most common and that is usually the slow path. The change @vinniefalco added looks good.  By the way, the reason to actually use the number in the fuzzer is to avoid the optimizer removing the code and exercise more code paths.

PR #203, opened 2020-08-24
UTF-8 validation is enabled for benchmarks

sdkrystian:


vinniefalco:
ooh nice commit message

vinniefalco:
could be shorter though :)

> benchmarks validate utf-8

PR #204, opened 2020-08-24
benchmarked libraries are not submodules

### LINKED ISSUE #70: Don't use submodules
bench should get nlohmann and rapidjson from a script not as a submodule

sdkrystian:
fix #70

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,0 @@
-[submodule "bench/lib/nlohmann"]
-	path = bench/lib/nlohmann
-	url = https://github.com/nlohmann/json.git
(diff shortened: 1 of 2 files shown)


vinniefalco:
What, you don't want to ship copies of RapidJSON and nlohmann's JSON with Boost?

sdkrystian:
LOL

sdkrystian:
@vinniefalco Should we add `/bench/lib` to the gitignore?

vinniefalco:
Yeah we should add it to gitignore if running the script would change the working set

vinniefalco:
I added it to .gitignore

PR #209, opened 2020-08-25
max_depth is specified with parse_options

### LINKED ISSUE #205: Put max depth in parse_options
it makes sense there, and then `parser` doesn't have to provide the member function. But then we need a way to query the parse options on the `parser` (and `basic_parser`).

sdkrystian:
fix #205

### DIFF
--- a/include/boost/json/detail/basic_parser.hpp
+++ b/include/boost/json/detail/basic_parser.hpp
@@ -189,9 +189,6 @@ class basic_parser
     number num_;
     error_code ec_;
     detail::stack st_;
-    std::size_t max_depth_ = 32;
-    // how many levels deeper the parser can go
-    std::size_t depth_ = max_depth_;
     detail::utf8_sequence seq_;
     unsigned u1_;
     unsigned u2_;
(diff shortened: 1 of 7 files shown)


include/boost/json/detail/basic_parser.hpp
```diff
@@ -189,16 +189,15 @@ class basic_parser
...
-    std::size_t depth_ = max_depth_;
     detail::utf8_sequence seq_;
     unsigned u1_;
     unsigned u2_;
     bool more_; // false for final buffer
     bool complete_ = false; // true on complete parse
     const char* end_;
     parse_options opt_;
+    // how many levels deeper the parser can go
+    std::size_t depth_ = opt_.max_depth;
```
vinniefalco, line 200: could call it `depth_remain_`. or not.

sdkrystian: Meh, not too important. If you want to rename it, go ahead.

vinniefalco: I don't feel strongly about it

PR #210, opened 2020-08-26
value_builder optimization

vinniefalco:
Has a small leak unfortunately


sdkrystian:
@vinniefalco Is the leak inherent to the approach? Or is it something we can fix cleanly.

vinniefalco:
nah just a Plain ole' bug. I fixed the leak in the "tidy" commit. However, now we have this "maybe used uninitialized" which has got me stumped, have a look?

sdkrystian:
Affirmative.

PR #219, opened 2020-08-28
Value stack

vinniefalco:
More improvements and refactoring


include/boost/json/parse_options.hpp
```diff
@@ -21,27 +21,45 @@ namespace json {
...
 {
-    /** Maximum parse depth
+    /** Maximum nesting level of arrays and objects.
         
-        The maximum number of nested structures
-        permitted within a JSON document.
+        This specifies the maximum number of nested
+        structures allowed while parsing a JSON. If
+        this limit is exceeded, the parser returns
+        an error immediately.
```
sdkrystian, line 36: I don't think we need to say "immediately".

include/boost/json/parser.hpp
```diff
@@ -121,22 +142,77 @@ class parser
...
+        @li The @ref memory_resource for the
+        implementation to use when it needs to
+        acquire temporary memory,
+
+        @li The @ref parse_options to use, which
+        can allow non-standard JSON extensions such
+        as comments or trailing commas, and
+
+        @li A caller-owned temporary buffer to use
+        first, before allocating using the memory
```
sdkrystian, line 157: Maybe:
> A caller-owned temporary buffer to use before allocating with the memory resource [...]

PR #256, opened 2020-09-03
test pr benchmarks

sdarwin:
Don't merge.
Testing the functionality of benchmark computation on pull requests. Processing takes an hour or so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,3 +59,4 @@ The design of the library also achieves these goals:
 Distributed under the Boost Software License, Version 1.0.
 (See accompanying file [LICENSE_1_0.txt](LICENSE_1_0.txt) or copy at
 https://www.boost.org/LICENSE_1_0.txt)
+ 


sdarwin:
Looks good. Closing PR.

vinniefalco:
I thought the image would be inlined in the message

sdarwin:
> I thought the image would be inlined in the message

I will what can be done about showing the images in a smaller window... Re-opening the issue.

sdkrystian:
Hm, the wild variance in the results might be my fault. For gcc it might be comparing against clang develop.... I'll check

sdarwin:
> Hm, the wild variance

I was also surprised by that.  But it wasn't an exact comparison because the develop branch changed since the pull request was submitted.   Further testing of the same commit with itself shows less variance.

vinniefalco:
is this pull request actionable?

sdarwin:
> is this pull request actionable?

It has become actionable, because it seems the build job is occasionally triggering for no reason.  The benchmarks ran, however there wasn't a new commit.  Investigating...

sdarwin:
Latest theory:  Although no new commits have been submitted to the PR, the "target" or "base" of the PR (which is the develop branch) has changed in the meantime, and that is causing the plugin to sometimes trigger rebuilds. 

@sdkrystian  -  benchmark computations have been revised to compare against the point of divergence from the develop branch instead of the newest commit.

PR #261, opened 2020-09-03
Add fuzzerbadge and improve documentation

### LINKED ISSUE #237: fuzzing badge
Per-commit fuzzing results should be available on a badge in the readme.md

pauldreik:
Added a fuzzer badge, Closes #237. I did not add a link for master, since that seems to not get updated by the cron job and I do not want to give the false impression that master is fuzzed.

Also, I saw a recent crash detected by the fuzzer job so I minimized it and documented how I did it.


pauldreik:
The crash was introduced in https://github.com/CPPAlliance/json/commit/c4dc2b2052e264a97e8e512ae2a5ce83814b797e

The CI job caught it.

vinniefalco:
This has conflicts and needs rebasing

vinniefalco:
Can we fuzz master too?

pauldreik:
> Can we fuzz master too?

I guess, but that would need some github action magic I don't know how to do even if I am sure it's easy.

Worse is, there is a risk of something I call corpus ping-pong. If master lags develop or the branches diverge, one will destroy the corpus for the other. Then we need to separate the corpora and that needs more of that unpleasant github action - bash script - bintray debugging. Of course it can be done, I just don't feel like it is worth the effort right now. I would rather wait until (if?) this library gets accepted into boost, and then get it onto oss fuzz instead which has a much more professional setup.

pauldreik:
> This has conflicts and needs rebasing

fixed

vinniefalco:
corpus ping-pong... lmao. OK

PR #268, opened 2020-09-04
Set parse_options from fuzzdata and invoke to_string()

pauldreik:
This should cover more of the code. I have run this locally, no crashes found.

### DIFF
--- a/fuzzing/fuzz_parser.cpp
+++ b/fuzzing/fuzz_parser.cpp
@@ -8,13 +8,14 @@
 //
 
 #include <boost/json/parser.hpp>
-
+#include <boost/json/parse_options.hpp>
 using namespace boost::json;
 
 bool
-fuzz_parser(string_view sv)
+fuzz_parser(parse_options popt,string_view sv)
 {
-    parser p;
+
+    parser p(storage_ptr{},popt);
     error_code ec;
 
     // This must be called once before parsing every new JSON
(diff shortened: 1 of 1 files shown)


fuzzing/fuzz_parser.cpp
```diff
@@ -39,14 +40,27 @@ fuzz_parser(string_view sv)
...
+        const uint8_t* data, size_t size)
 {
+    if(size<1)
+        return 0;
+
+    parse_options popt;
+    popt.allow_comments=!!(data[0]&0x1);
+    popt.allow_trailing_commas=!!(data[0]&0x2);
+    popt.allow_invalid_utf8=!!(data[0]&0x4);
+    popt.max_depth= (data[0]>>3);
```
vinniefalco, line 53: hey that's really cool!!!

pauldreik: Yep, only cool kids do bit shifts and stuff :-)

vinniefalco: LOL no... I mean, the mechanism for prodding the fuzzing tool to fill out the parse options

pauldreik: Why are you not impressed with my coding skillz? I can do bitwise or, too!

On a more serious note, yes it is awesome to fill out input data from the fuzzer stream. Have a look at the curl fuzzer, which sets flags and then cuts different parts of the input data as the (fake) server response.

PR #277, opened 2020-09-05
Drop bintray, use github action cache

pauldreik:
@Flamefire please have a look

### DIFF
--- a/.github/workflows/run_fuzzer.yml
+++ b/.github/workflows/run_fuzzer.yml
@@ -1,9 +1,5 @@
 name: fuzz
 
-env:
-  BINTRAYUSER: vinniefalco
-  BINTRAYORG: cppalliance
-
 on:
   push:
   pull_request:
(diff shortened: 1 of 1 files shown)


pauldreik:
When run for the first time, it correctly stores the cache and it is found on the next run (at least when running on the same branch).
The problem comes at the end of second run. Because the cached item already exists, it won't be updated even if the file has changed.

I can not find how to evict the cache manually.

If that eviction feature would have existed, using github action cache could have worked for storing the corpus.

@jonathanmetzman maybe you know more than me on how this could work?

Flamefire:
This  should work in principle. See my test repo:  https://github.com/Flamefire/gha_tests
Check the actions tab for the 2 "test" commit runs and expand " Run value=$(cat cacheFolder/foo.txt)" which shows how the cached value is increased with every run

pauldreik:
> This should work in principle. See my test repo: https://github.com/Flamefire/gha_tests
> Check the actions tab for the 2 "test" commit runs and expand " Run value=$(cat cacheFolder/foo.txt)" which shows how the cached value is increased with every run

I see that it works in your repo, thanks for the demo! I think I know why it did not work for me:
I used the "rerun workflow" button and not pushing new changes. The new run gets the same id. On the second run, I get "Cache hit occurred on the primary key corpus, not saving cache." 

That means the suggested approach would work, but not on reruns. I will do an experiment with a random key instead and see what happens.

pauldreik:
Ok, I found an ugly hack to get this to work. Using a symlink to /proc/self/stat placed within the repo, then using the github action supplied function hashFiles() on the symlink.

pauldreik:
I think this is ready for review/merge now.

pauldreik:
Thank you @Flamefire  for your help!

Flamefire:
> I used the "rerun workflow" button and not pushing new changes. The new run gets the same id. On the second run, I get "Cache hit occurred on the primary key corpus, not saving cache."

I'd consider that a feature and keep that. If you rerun you often like to see the same failure again. Besides that I very rarely use the "rerun" button and just trigger the CI with a new commit (you can even amend the last commit without changing anything and force push) so I wouldn't include such hacks for such a minor use case.
Just for your consideration.

pauldreik:
> > I used the "rerun workflow" button and not pushing new changes. The new run gets the same id. On the second run, I get "Cache hit occurred on the primary key corpus, not saving cache."
> 
> I'd consider that a feature and keep that. If you rerun you often like to see the same failure again. Besides that I very rarely use the "rerun" button and just trigger the CI with a new commit (you can even amend the last commit without changing anything and force push) so I wouldn't include such hacks for such a minor use case.
> Just for your consideration.

This is for a fuzz corpus so it is only to speed up the fuzzing. If something interesting happens, it ends up in one of the files caught by the artifacts job. Hence there is no need to rerun with the exact same corpus.

Regarding rerun, it is a very handy feature if one wants to manually add some extra fuzzing effort, for instance when wondering why a bug is not caught by the fuzzer.

For this particular CI job, this is exactly how I want it :-)

.github/workflows/proc_self_stat
```diff
@@ -0,0 +1 @@
+/proc/self/stat
```
vinniefalco, line 1: This is some of your best work right here

vinniefalco: This file is giving me trouble on Windows. When I check out the branch, the file shows up as modified in my working set. WIndows doesn't support symlinks like this. I"m not sure what to do.

pauldreik: That means a large percentage of users will have the same problem, leading to questions even if I am sure there is some symlink setting in windows git somewhere which might fix this.

Well, I guess that means I will drop that nice hack.

pauldreik: ok, pushed a new version with the hack dropped.

vinniefalco:
Merged thanks

PR #279, opened 2020-09-05
fix 32 bit warnings for prime table

### LINKED ISSUE #270: prime table operation is not 32-bit safe
This line causes a warning:
https://github.com/CPPAlliance/json/blob/adacd6670cd92ef5b8e2890550a7a6171e86d854/include/boost/json/impl/object.ipp#L512

The way it is written, it looks like we can get a `*prime` that exceeds the max unsigned 32-bit integer. Then the assignment will underflow.

sdkrystian:
fix #270

@vinniefalco Could you test this on 32 bit?


vinniefalco:
The problem is bigger than this, so closing for now

PR #289, opened 2020-09-06
fix accidental space in yaml file

pauldreik:
sorry


vinniefalco:
does yaml care?

pauldreik:
> does yaml care?

I wish not...

PR #291, opened 2020-09-06
Max size

vinniefalco:
fixes the problem with limits


sdkrystian (commented):
Looks good. Though, we should just define `max_size` as `soft_limit`.

vinniefalco:
> max_size as soft_limit.

That is something we can explore in the future, but note that there is a circular dependency loop of include files, so writing `sizeof(key_value_pair)` in object.hpp will be challenging.

PR #296, opened 2020-09-07
Note that string doesn't validate UTF-8

### LINKED ISSUE #252: string docs could be more clear about utf8
The string javadoc should make it clear that it is the caller's responsibility to ensure only valid utf-8 is placed in the string.

sdkrystian:
fix #252


vinniefalco:
Merged, thanks

PR #297, opened 2020-09-07
serializer refactor

vinniefalco:
* serialize() replaces to_string()
    * serialize.hpp replaces to_string.hpp
    * serializer works on all types
    * serialize, operator<< work on all types
    * reset() takes pointer arguments
    * resetting constructor removed


sdkrystian:
Looks good save for the fuzzer failure. Maybe we should add support for integers?

PR #298, opened 2020-09-07
Handler specifies structure and string limits

sdkrystian:


example/validate.cpp
```diff
@@ -26,60 +26,64 @@ using namespace boost::json;
 
 //[example_validate
 
-bool
-validate( string_view s )
+// The null parser discards all the data
+class null_parser
```
vinniefalco, line 30: Can't put it in the function?

include/boost/json/detail/array_impl.hpp
```diff
@@ -54,10 +54,7 @@ class array_impl
     static
     constexpr
     std::size_t
-    max_size() noexcept
-    {
-        return BOOST_JSON_MAX_STRUCTURED_SIZE;
-    }
+    max_size() noexcept;
```
vinniefalco, line 57: are constexpr functions inline by default?

include/boost/json/parser.hpp
```diff
@@ -107,6 +107,26 @@ class parser
 {
     struct handler
     {
+        constexpr
+        static
+        std::size_t
+        max_object_size = object::max_size();
```
vinniefalco, line 113: Consider
```
constexpr static std::size_t
    max_object_size = object::max_size();
```
since it is private

include/boost/json/detail/array_impl.hpp
sdkrystian: Yup

sdkrystian:
@sdarwin I think I see the problem now -- we should be comparing against the commit prior to the first commit on a PR instead of comparing to the current develop

include/boost/json/basic_parser.hpp
```diff
@@ -967,10 +967,20 @@ parse_unescaped(const char* p)
     cs = detail::count_valid<AllowBadUTF8>(
         cs.begin(), cs.end());
     std::size_t size = cs.used(start);
-    if(BOOST_JSON_UNLIKELY(size > 
-        BOOST_JSON_MAX_STRING_SIZE - total))
-        return fail(cs.begin(), IsKey ? 
-            error::key_too_large : error::string_too_large);
+    if(IsKey)
+    {
```
vinniefalco, line 971:
```
BOOST_ASSERT(total <= Handler::max_key_size_)
```

PR #299, opened 2020-09-07
default_resource singleton is a static data member

sdkrystian:


vinniefalco:
needs tests for order-of-init

vinniefalco (changes requested):
still needs tests for order-of-init

include/boost/json/impl/storage_ptr.ipp
```diff
@@ -15,6 +15,16 @@
 BOOST_JSON_NS_BEGIN
 
 // KRYSTIAN NOTE: we need P1077
+#ifdef __clang__
```
vinniefalco, line 18: P1077 won't help us until what, C++23? Or C++26? Or never? And even then, we still have to support older versions of C++. Therefore, P1077 will never be of any help.

include/boost/json/impl/storage_ptr.ipp
```diff
@@ -0,0 +1,52 @@
...
+};
+
+template<class T>
+struct singleton
+{
+    static no_destroy resource_;
+};
+
+template<class T>
+no_destroy singleton<T>::resource_;
```
vinniefalco, line 50: You are injecting all these into the `boost::json` namespace, I think they should go in `boost::json::detail`?

include/boost/json/storage_ptr.hpp
```diff
@@ -173,7 +169,8 @@ class storage_ptr
         No-throw guarantee.
     */
     storage_ptr() noexcept
-        : i_(0)
+        : i_(reinterpret_cast<
+            std::uintptr_t>(get_default()))
```
vinniefalco, line 173: this is GUD

include/boost/json/storage_ptr.hpp
```diff
@@ -337,8 +335,9 @@ class storage_ptr
         storage_ptr&& other) noexcept
     {
         release();
-        i_ = detail::exchange(
-            other.i_, 0);
+        i_ = detail::exchange(other.i_,
+            reinterpret_cast<
+                std::uintptr_t>(get_default()));
```
vinniefalco, line 340: Should `get_default` maybe return `std::uintptr_t`? It would clean up these casts.

include/boost/json/impl/storage_ptr.ipp
```diff
@@ -0,0 +1,52 @@
...
+
+memory_resource*
+storage_ptr::
+get_default() noexcept
+{
+    return &singleton<void>::
+        resource_.resource_;
```
vinniefalco, line 57: `resource_.resource_` yuck

include/boost/json/impl/storage_ptr.ipp
sdkrystian: I'll just move this to `default_resource`

include/boost/json/storage_ptr.hpp
sdkrystian: Yup, did that

include/boost/json/impl/storage_ptr.ipp
sdkrystian: Fixed

PR #300, opened 2020-09-07
Add detail::unreachable

sdkrystian:


include/boost/json/detail/config.hpp
```diff
@@ -209,6 +211,28 @@
...
+#pragma clang diagnostic ignored "-Winvalid-noreturn"
+#elif defined(__GNUC__)
+#pragma gcc diagnostic push
+#pragma gcc diagnostic ignored "-Winvalid-noreturn"
+#endif
+
+BOOST_NORETURN
+BOOST_FORCEINLINE
+void
+unreachable()
```
vinniefalco, line 225: doesn't this need to be marked `inline`?

sdkrystian: `BOOST_FORCEINLINE` does that for us

sdkrystian: `BOOST_FORCEINLINE` does that for us

PR #301, opened 2020-09-07
Fix serialize fuzzing failure

sdkrystian:


vinniefalco:
This should be fixed by the std::string change, sorry!

PR #310, opened 2020-09-10
Binary size reductions

sdkrystian:


include/boost/json/detail/config.hpp
```diff
@@ -198,6 +198,16 @@
...
 
+#ifndef BOOST_JSON_UNREACHABLE
+# if defined(__GNUC__) || defined(__clang__)
+#  define BOOST_JSON_UNREACHABLE() __builtin_unreachable()
+# elif defined(_MSC_VER)
+#  define BOOST_JSON_UNREACHABLE() __assume(0)
+# else // infinite loops have undefined behavior
+#  define BOOST_JSON_UNREACHABLE() do { } while(1)
+# endif
+#endif
```
vinniefalco, line 209: @pdimov is this correct?

pdimov: The [[noreturn]] fallback is probably better than an infinite loop.

vinniefalco:
I think we should just have the commit with the `parse_comment` trimmed down, and worry about getting the unreachable() thing right later. This would give an immediate savings, with the promise of more down the line.

vinniefalco:
Merged

PR #318, opened 2020-09-11
Add visit

### LINKED ISSUE #207: value::visit() function?
This would be easy to implement, do we need it? Should we just make it an example?

vinniefalco:
fix #207

### DIFF
--- a/include/boost/json/visit.hpp
+++ b/include/boost/json/visit.hpp
@@ -1,0 +1,50 @@
+//
+// Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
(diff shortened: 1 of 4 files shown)


include/boost/json/visit.hpp
```diff
@@ -0,0 +1,50 @@
...
+    @param v The visitation function to invoke
+
+    @param jv The value to visit.
+*/
+template<class Visitor>
+auto
+visit(
+    Visitor&& v,
+    value&& jv) -> decltype(
+        std::declval<Visitor>()(nullptr))
```
sdkrystian, line 32: Do `std::forward<Visitor>(v)`

include/boost/json/visit.hpp
```diff
@@ -0,0 +1,50 @@
...
+{
+    switch(jv.kind())
+    {
+    default: // unreachable()?
+    case kind::array:   return v(jv.get_array());
```
sdkrystian, line 37: Same thing here

vinniefalco:
Merged

PR #321, opened 2020-09-12
Fix alignment bug in ~key_value_pair

sdkrystian:


include/boost/json/impl/value.ipp
```diff
@@ -455,7 +455,7 @@ key_value_pair::
     if(sp.is_not_counted_and_deallocate_is_trivial())
         return;
     sp->deallocate(const_cast<char*>(key_),
-        len_ + 1, 1);
+        len_ + 1);
```
vinniefalco, line 458: Wouldn't it be better to change the alignment in the call to `allocate` to be 1?

PR #322, opened 2020-09-12
Review docs

sdkrystian:


include/boost/json/value.hpp
```diff
@@ -2215,7 +2215,7 @@ class value
 
     /** Return the underlying `std::uint64_t`, or throw an exception.
 
-        If @ref is_int64() is `true`, returns
+        If @ref is_uint64() is `true`, returns
```
vinniefalco, line 2218: `is_uint64` returns a pointer not a bool, same with the others

sdkrystian: How should it be worded then? "If is_uint64_t() returns a non-null pointer"?

vinniefalco: actually we are going to roll back the pointer change, stay tuned

vinniefalco:
This needs rebasing, since I refactored the `is_array` interfaces

vinniefalco:
This was never rebased and thus cannot be merged.

PR #324, opened 2020-09-12
Fix limits test on compilers that default to C++98.

qis:
This is a fix for out-of-source boost build with tests enabled on compilers that default to C++98 (e.g. Apple Clang) as discussed on the cpplang slack in #boost-json.


vinniefalco:
Merged, thanks!

PR #329, opened 2020-09-13
Various

vinniefalco:


include/boost/json/parser.hpp
```diff
@@ -29,16 +29,17 @@ BOOST_JSON_NS_BEGIN
...
-    must be called, optionally passing the storage
-    pointer to be used by the @ref value container into
-    which the parsed results are stored. After the
-    parse is started, call the @ref write function
-    to provide buffers of characters of the JSON.
-    When there are no more buffers, call @ref finish.
-    The parse is complete when the function @ref done
-    returns `true`, or when a non-successful error
-    code is returned.
+    Parsing for a new JSON may begin the parser is
```
thats-bonkers, line 32: This sentence looks broken... possibly meant to say 'when the parser is'

include/boost/json/parser.hpp
```diff
@@ -156,127 +157,137 @@ class parser
...
+        `buffer` extends until the parser is destroyed.
     */
     BOOST_JSON_DECL
     parser(
         storage_ptr sp,
         parse_options const& opt,
-        unsigned char* temp_buffer,
-        std::size_t temp_size) noexcept;
+        unsigned char* buffer,
+        std::size_t size) noexcept;
```
thats-bonkers, line 193: Have you considered using a string_view here... or is it just a matter of making calls using `std::byte*` less clunky?

include/boost/json/parser.hpp
vinniefalco: This will be fixed elsewhere

include/boost/json/parser.hpp
sdkrystian: `string_view` is non-mutable

thats-bonkers: Oh, my bad, I made the wrong assumption here... I guess that a span-like type could be useful here then, but if you don't have any such construct in this project, it doesn't pay off to bring it from another dependency.

include/boost/json/parser.hpp
thats-bonkers: Ack

thats-bonkers (approved):

PR #359, opened 2020-09-18
value::emplace takes arguments

vinniefalco:
This adds arguments to value::emplace


vinniefalco:
Closing this because the algorithm is less efficient than using the noexcept emplace with no args followed by a `resize` (e.g. for arrays)

PR #361, opened 2020-09-18
Converted multiple headers into single header 

siddharth25pandey:
As in bench directory,in bench.cpp file. There are unnecessary multiple header file which are already present in the **bits/stdc++.h**
That can reduce the space as well code execution time.


vinniefalco:
There are some problems with this pull request:

1. Pull requests should be made against the **develop** branch, not the **master** branch

2. `<bits/stdc++.h>` is implementation-specific and not a standard include file

3. Your commit contains unnecessary whitespace changes which make reviewing more difficult

siddharth25pandey:
@vinniefalco  <bits/stdc++.h> is implementation-specific but it already have headers like algorithm,vector,cstdio,etc.
That will reduce the time cost as well as memory consumption.Please have a insider look
Okay i will make PR against develop branch.
Thanks!

siddharth25pandey:
@vinniefalco Please give confirmation regarding this. So, that I will create a new PR against develop branch.
Thanks!

Olipro:
> 
> 
> @vinniefalco <bits/stdc++.h> is implementation-specific but it already have headers like algorithm,vector,cstdio,etc.
> That will reduce the time cost as well as memory consumption.Please have a insider look
> Okay i will make PR against develop branch.
> Thanks!

It will also make it entirely unusable on platforms which do not utilise the GCC standard C++ library. This is a non-starter for that reason, and for the fact it is also not conformant to standard, portable C++.

PR #371, opened 2020-09-20
Refactor array

vinniefalco:
Less code, faster


include/boost/json/impl/array.ipp
```diff
@@ -119,25 +240,26 @@ array(
...
+        t_ = &empty_;
+        return;
+    }
+    t_ = table::allocate(
+        count, sp_);
+    t_->size = count;
+    auto p = data();
+    do
+    {
+        ::new(p++) value(sp_);
```
sdkrystian, line 254: Something that has applications outside of just `array` (you might not like this because it would be private interface): we could just `memcpy` the `storage_ptr` and at the end just have a single branch which checks `is_counted`, and if that returns `true` we increment the ref-count by `count`. It would kill a branch in each construction (thus optimizing the non-counted case) and we would only have to do one atomic add instead of `count` (atomic adds are expensive -- `lock add QWORD PTR[rax], 1` takes 3x as many cycles as `add QWORD PTR[rax], 1`), which would optimize the counted case.

sdkrystian: This would especially help in `value_stack`. Since we don't do deep copies, we can check `is_counted` and increment the ref-counter only when `push_array`, `push_object` and `release` are called.

include/boost/json/impl/array.hpp
```diff
@@ -18,56 +18,167 @@
...
+struct alignas(alignof(value))
+    array::table
 {
-    array& self_;
+    std::uint32_t size;
+    std::uint32_t capacity;
 
-public:
-    bool commit = false;
+    char pad_[sizeof(value) - (
```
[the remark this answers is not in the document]
vinniefalco: yeah that's the theory but my static asserts fail without it

vinniefalco:
Merged

PR #414, opened 2020-09-25
Use switch instead of jump table

sdkrystian:


sdkrystian:
Addressed by c547ca9d21c55bd045fcc13455c8dd311c7619ce

PR #415, opened 2020-09-25
tag_invoke exposition

### LINKED ISSUE #411: Refer to P1895 for tag_invoke
> I would make a note in the Value conversion mechanism referencing
> P1895 on `tag_invoke` so users who don't know this paper don't
> stare at the name wondering 'why?'

sdkrystian:
fix #411

### DIFF
--- a/doc/qbk/03_06_conversion.qbk
+++ b/doc/qbk/03_06_conversion.qbk
@@ -18,8 +18,47 @@ The conversion in the opposite direction is done with __value_to__.
 
 [snippet_conv_1]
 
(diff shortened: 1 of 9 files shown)


vinniefalco:
Better

vinniefalco:
This fails to build.

vinniefalco:
One of the tests throws an exception.

vinniefalco (changes requested):
This fails to build on some configurations, and a test fails.

vinniefalco:
Merged, thanks

PR #436, opened 2020-10-01
Analyze-bloat workflow

djarek:
This workflow uses Bloaty McBloatface to generate binary bloat data. Additionally, if the event is a PR, a diff between PR HEAD and base will be generated.


vinniefalco:
oh, this is very nice

vinniefalco:
Merged, thanks

PR #446, opened 2020-10-08
Use relaxed increment and acquire_release decrement for atomic reference counts

maximilianriemensberger:
Implements #446


vinniefalco:
It looks like your local git credentials are not set, because the commit is not attached to a GitHub account, would you mind doing `git config` and setting your email and name?

vinniefalco:
merged, thanks!

PR #447, opened 2020-10-09
Test for esoteric double parse accuracy

madmongo1:


vinniefalco:
This looks like it is failing on Azure?

madmongo1:
Yep this is the issue I mentioned on slack. Vs2017 produces a different result to all other compilers by 2ulp for that case.
Not sure where we go from here.

vinniefalco:
Can we increase the tolerance just for VS2017 and earlier?

PR #455, opened 2020-10-22
Add hash example

madmongo1:
documentation embedding will need a review

### DIFF
--- a/example/hash.cpp
+++ b/example/hash.cpp
@@ -1,0 +1,166 @@
+//
+// Copyright (c) 2020 Richard Hodges (hodges.r@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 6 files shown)


example/hash.cpp
```diff
@@ -0,0 +1,166 @@
...
+    std::size_t seed = 0;
+    for (auto &x : arg)
+        hash_combine(seed, x);
+    return seed;
+}
+
+std::size_t hash< json::object >::operator()(json::object const &arg) const
+{
+    std::size_t                                   seed = 0;
+    thread_local std::vector< json::string_view > keys;
```
pdimov, line 92: No need for thread_local on a local variable.

vinniefalco: I think he meant to wrtite `thread_local static`

grisumbras:
Implemented as a library feature

PR #456, opened 2020-10-22
Fix visit for const values

madmongo1:


test/visit.cpp
```diff
@@ -32,21 +46,97 @@ class visit_test
     };
 
     void
-    testVisit()
+    testVisitMutable()
+    {
+        {
+            json::value v = value(nullptr);
+            BOOST_TEST(visit(check_mutable{kind::null}, v));
```
vinniefalco, line 53: Why did we bloat the number of lines?

madmongo1: Because we have to prove the correct path is taken when passing lvalues.
Previously the test was passing rvalues.

vinniefalco: Yepp... I see what you mean now :)

PR #460, opened 2020-10-26
Refactor CI

sdkrystian:


include/boost/json/value.hpp
```diff
@@ -2479,12 +2480,13 @@ class value
             ! std::is_same<T, bool>::value,
                 T>::type
     {
+        using limits = std::numeric_limits<T>;
```
vinniefalco, line 2483: Is the alias necessary for fixing the warnings?

vinniefalco: If you are going to change it use `DBL_MIN` and `DBL_MAX` or whatever those macros are called

sdkrystian: The alias isn't necessary, I just added it for convenience. We can use the macros instead

vinniefalco:
Merged, thanks

PR #462, opened 2020-10-31
Prerelease versions do not use -Werror

sdkrystian:


vinniefalco:
Yeah this is good thanks

PR #472, opened 2020-11-24
fix drone asan build

cppalliance-drone:
A few of the jobs in drone were updated from Ubuntu 16 to Ubuntu 18.    Let's try switching asan back to Ubuntu 16.


cppalliance-drone:
Build is passing https://drone.cpp.al/boostorg/json/3

PR #487, opened 2021-01-24
Add function overload throw_exception

### LINKED ISSUE #398: throw_exception in separate namespace in standalone
When exceptions are not disabled, the provided overload of `throw_exception` for standalone should be in the `json::detail` namespace to avoid collisions when linking with other libraries.

Appendme:
There is no function overload for boost version >= 1.73.0
close #398


grisumbras:
Since #398 is fixed, we can close this.

PR #492, opened 2021-01-26
Drop support below 1 75

### LINKED ISSUE #490: Boost versions below 1.75 are not supported
The check for
```
#if BOOST_VERSION >= 107300
```
in except.ipp should be changed to
```
#if ! defined(BOOST_JSON_STANDALONE)
```
because Boost.JSON no longer supports versions of Boost below 1.75. The check for 1.73 was added only before the first official Boost release, so that users could test with their OS-provided Boost.

grisumbras:
Fixes #490 

Replaces checks of Boost version with checks for whether the library is used standalone. Replaces BOOST_CURRENT_LOCATION with a custom macro.


include/boost/json/detail/except.hpp
```diff
@@ -25,9 +25,10 @@ namespace detail {
 
 // VFALCO we are supporting Boost 1.67 because it is in a lot of distros
```
vinniefalco, line 23: This comment is no longer applicable

vinniefalco:
Commit message should read "Add BOOST_JSON_SOURCE_POS"

vinniefalco:
Commit message should read "Don't check Boost version"

grisumbras:
The latest update makes it so that standalone_with_boost only tested when everything else is tested _not in standalone mode_. This is due to the fact, that the test needs Boost present and standalone CI tests don't have it.

test/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ target_compile_definitions(limits PRIVATE
 if(BOOST_JSON_STANDALONE)
     target_compile_definitions(limits PRIVATE BOOST_JSON_STANDALONE)
     target_compile_features(limits PRIVATE cxx_std_17)
+    list(FILTER BOOST_JSON_TESTS_FILES EXCLUDE REGEX "standalone_with_boost.cpp")
```
vinniefalco, line 36: you nasty boy

grisumbras:
> Error: retrieving gpg key timed out.

This appears to be a spurious CI error

PR #497, opened 2021-02-05
Fix (lack of) building without C++11 support

grisumbras:


grisumbras:
This should fix Boost builds without C++11 support

vinniefalco:
Suggested commit message: "Builds prior to C++11 are no-ops"

grisumbras:
Do you think there's a need for CI test for no-op build with cxxstd=03?

vinniefalco:
Please change "Adds back GCC 4.8 support" to "gcc-4.8 is supported"

vinniefalco:
> Do you think there's a need for CI test for no-op build with cxxstd=03?

Up to you

vinniefalco:
Can we please change the commit message to "gcc-4.8 is supported" ? Use an interactive rebase (`git rebase -i`) then force-push the branch.

grisumbras:
Which commit?

vinniefalco:
This one: 
image

PR #498, opened 2021-02-10
Comment out microsoft windows drone build

sdarwin:
The CI script to test on the Windows platform may need to be redesigned, based on recent changes in boost build.


vinniefalco:
Obsolete

PR #514, opened 2021-03-03
Fix some typo error in value, string and conversion documents.

expertcxxmoon:
Found some typo while reading the doc then submit patch to fix the typo.

Signed-off-by: Liang Yan


vinniefalco:
Ah yes good eyes!

vinniefalco:
Merged, thank you

PR #515, opened 2021-03-03
Value to tuple

### LINKED ISSUE #508: value_to tuple/pair
`value_to` should work for tuple-like and pair-like types. This interacts with #466 because `array` is tuple-like and needs to be handled as a special case.

grisumbras:
Fix #508

### DIFF
--- a/include/boost/json/detail/index_sequence.hpp
+++ b/include/boost/json/detail/index_sequence.hpp
@@ -1,0 +1,44 @@
+//
+// Copyright (c) 2021 Dmitry Arkhipov (grisumbras@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/json
+//
+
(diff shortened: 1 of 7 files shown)


include/boost/json/detail/index_sequence.hpp
```diff
@@ -0,0 +1,44 @@
...
+# include <type_traits>
+#endif
+
+BOOST_JSON_NS_BEGIN
+namespace detail {
+
+#if ! defined(BOOST_JSON_STANDALONE)
+
+template <std::size_t... Is>
+using index_sequence = boost::mp11::index_sequence<Is...>;
```
vinniefalco, line 26: Very nice

vinniefalco:
I really don't have the skills to review stuff related to `value_to` and `value_from`. @pdimov ?

pdimov:
If it passes msvc-14.0 CI it's probably good.

pdimov:
Looks good to me. To repeat what I said on Slack, now that tuple-likes are supported it looks like general maps (where the key is not string-like) would also start working (being treated as sequences of pairs.) If they do work, we need to have a test for it. If they don't, we might want to look into why.

PR #519, opened 2021-03-04
Value from char const*

### LINKED ISSUE #507: value_from doesn't support char const*
`value_from( "hello" )` should probably work...

grisumbras:
Fixes #507.

### DIFF
--- a/test/value_from.cpp
+++ b/test/value_from.cpp
@@ -127,19 +127,27 @@ check(
     }
 }
 
+BOOST_JSON_NS_BEGIN
+
+namespace {
+
+template<class T>
+static
+void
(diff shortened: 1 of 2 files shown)


test/value_from.cpp
```diff
@@ -173,7 +181,10 @@ class value_from_test
     {
         // value_from supports every value constructor
 
-        testValueCtor<value const&>();
+        testValueCtor<value>();
+
+        char const* s = "5";
+        testValueCtor(s);
```
vinniefalco, line 187: It looks like these tests are incomplete? `testValueCtor` is only called from 1 or 2 places. I'm not sure what is supposed to be happening here.

Issue #521, opened 2021-03-05
Add std::hash specializations


doganulus:
It might be nice to add std::hash support for `boost::json` types to enable their use in associative containers as keys. Currently the library only provides std::hash support for `boost::json::string`. For the full support, std::hash specializations for other constructs should be added.

```cpp
namespace std {

template <>
struct hash<boost::json::array> {
  size_t operator()(const boost::json::array& jv) const;
};

template <>
struct hash<boost::json::object> {
  size_t operator()(const boost::json::object& jv) const;
};

template <>
struct hash<boost::json::value> {
  size_t operator()(const boost::json::value& jv) const;
};

}  // namespace std
```

Then, in my implementation, I noticed that `boost::json::object` preserves the insertion order when iterating over `key_value_pair`s as my test case below failed initially.

```cpp
TEST_CASE("Boost Json Object Hash") {
  auto r1 = std::hash<boost::json::value>{}(
      boost::json::value({{"a", 11}, {"b", 12}}));
  auto r2 = std::hash<boost::json::value>{}(
      boost::json::value({{"b", 12}, {"a", 11}}));
  CHECK_EQ(r1, r2);
}
```
Currently, when calculating hashes for `boost::json::object`s, I additionally sort pairs with respect to keys beforehand but wonder if this feature can be implemented better and added to the library. 

Thank you!

vinniefalco:
hmm... `std::hash` is not defined for `vector`,  `map`, or `unordered_map`. However it is defined for `variant`.

As you noted, `json::object` is particularly interesting because it preserves order. An implementation of `std::hash<json::object>` would need to sort the keys every time it calculates the hash. We can't put this in the object itself because it would add overhead to everyone. There's no good choices here.

You can still use `json` containers in associative containers by defining your own hash function, as you noted.

doganulus:
Now I think of that a commutative hash function (e.g. simple xor'ing) would work for `json::object`. 

I am not sure about the quality of hashing as I do not have much experience with hash functions.

doganulus:
Here is my implementation. Open to suggestions.

```cpp
inline std::size_t combine(std::size_t seed, std::size_t h) {
  seed ^= h + 0x9e3779b9 + (seed << 6U) + (seed >> 2U);
  return seed;
}

inline std::size_t combine_comm(std::size_t x, std::size_t y) { return x ^ y; }

namespace std {

size_t hash<boost::json::array>::operator()(
    const boost::json::array& jv) const {
  auto seed =
[...]

PR #522, opened 2021-03-06
Add std::hash specializations for json types

doganulus:
As discussed for the issue #521, this pull request

* Add std::hash specializations for `json::array`, `json::object`, and  `json::value`.
* Add a new header `boost/json/detail/hash_combine.hpp` for inline hash functions.
* Add basic unit tests


include/boost/json/impl/array.ipp
```diff
@@ -754,4 +755,27 @@ equal(
...
+//----------------------------------------------------------
+
+namespace std {
+std::size_t 
+hash<::boost::json::array>::operator()(
+    ::boost::json::array const& ja) const noexcept 
+{
+  std::size_t seed = ja.size();
+  for (const auto& jv : ja) {
+    seed = boost::json::detail::hash_combine(
```
vinniefalco, line 771: `::boost::json::detail::hash_combine` ?

include/boost/json/impl/object.ipp
```diff
@@ -822,4 +823,30 @@ destroy(
...
+//----------------------------------------------------------
+
+namespace std {
+std::size_t 
+hash<::boost::json::object>::operator()(
+    ::boost::json::object const& jo) const noexcept 
+{
+    std::size_t seed = jo.size();
+    for (const auto& kv_pair : jo) {
+        const auto hk = std::hash<boost::json::string>{}(kv_pair.key());
```
vinniefalco, line 839: This constructs a temporary `json::string` from `key()` (which is a `string_view`). I don't think that's what you intended :)

include/boost/json/impl/value.ipp
```diff
@@ -494,4 +495,54 @@ key_value_pair(
...
+// std::hash specialization
+//
+//----------------------------------------------------------
+
+namespace std {
+std::size_t 
+hash<::boost::json::value>::operator()(
+    ::boost::json::value const& jv) const noexcept 
+{
+  std::size_t seed = static_cast<std::size_t>(jv.kind());
```
vinniefalco, line 509: Should `hash( jv )` == `hash( jv.get_array() )` if `jv.is_array()==true`? With the current code, they will not be equal. I'm not sure if they should be equal or not. @pdimov ?

test/value.cpp
```diff
@@ -2087,6 +2087,38 @@ class value_test
...
+        BOOST_TEST(hnull != hfalse);
+        BOOST_TEST(htrue != hfalse);
+        BOOST_TEST(hi0 == hu0);
+        BOOST_TEST(hi0 != hd0);
+
+        BOOST_TEST(hobj0 == hobj1);
+        BOOST_TEST(hobj0 != hobj2);
+
+        BOOST_TEST(harr0 == harr1);
+        BOOST_TEST(harr1 != harr2);
```
vinniefalco, line 2118: Nice test coverage, but lets be honest here - there would have been a small but non zero chance that these hashes collided :)

test/value.cpp
```diff
@@ -2108,6 +2140,7 @@ class value_test
         testStdConstruction();
         testInitList();
         testEquality();
+        testHash();
```
vinniefalco, line 2143: array.cpp and object.cpp each need their own small `testHash()` to exercise the corresponding hash function.

vinniefalco:
Couple of comments but this looks great, thanks !

include/boost/json/impl/value.ipp
pdimov: That's a change I requested. Variants typically hash the kind. E.g. the hash of variant<string, string> when it holds the same string as alternative 0 or alternative 1 will differ (which matches what operator== does.)

include/boost/json/impl/array.ipp
pdimov: I don't think there's any need to open `namespace std` here.

include/boost/json/impl/object.ipp
pdimov: Unfortunately std::hash<std::string_view> is C++17. :-)

include/boost/json/impl/value.ipp
pdimov: But, as I also said, this as written doesn't match the behavior of value::operator==, where int64 and uint64 are considered equal when they hold the same value.

pdimov: Scratch that, the code further on overrides the seed in the uint64 case. :-) However it assumes that std::hash<int64_t> and std::hash<std::uint64_t> produce the same value, which is not guaranteed.

test/value.cpp
pdimov: 2^-32 isn't bad.

I'd have written this differently - with a helper function that when v1 == v2 checks that the hashes match, and when v1 != v2 checks that they differ, so as to test the correspondence between op== and the hash. Or with two helper functions, one checking the equal case and one checking the unequal case.

Or I suppose one can just check equality inline.

Things like 0 != false, {} != 0, [] != 0, {} != [], [] != [[]], etc should be checked, too. The above doesn't check across kinds (except for null).

include/boost/json/impl/value.ipp
vinniefalco: Should `operator==` always return false when comparing int64 to uint64?

pdimov: I don't think there's any pressing need to change the current equality behavior at this time.

include/boost/json/impl/object.ipp
vinniefalco: Use
```
        auto const hk = ::boost::json::detail::digest(
            kv_pair.key().data(), kv_pair.key().size(), 0);
``

File: include/boost/json/impl/array.ipp (hunk 1 above)
Comment (doganulus) [reply include/boost/json/impl/array.ipp:771]: Once a user had reported me a GCC bug (affecting from GCC4.8 to 7) related to such specializations:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56480

The quick solution for that is to open the `namespace std`.

File: test/value.cpp (hunk 4 above)
Comment (doganulus) [reply test/value.cpp:2118]: Will update tests accordingly.

File: include/boost/json/impl/array.ipp (hunk 1 above)
Comment (pdimov) [reply include/boost/json/impl/array.ipp:771]: That's not the same. You're defining the member function here, not specializing the class template.

File: include/boost/json/impl/value.ipp (hunk 3 above)
Comment (doganulus) [reply include/boost/json/impl/value.ipp:509]: Question: What is the opinion regarding the equality check with different types (including primitives)? 

For example, these expressions all evaluate to `true`:

```cpp
value(nullptr) == nullptr
value(true) == true
value(false) == false
value(13) == 13
value({1,2,3}) == array({1,2,3})
value({{"a", 1}, {"b",2}}) == object({{"a", 1}, {"b",2}}) 
```
as right hand sides are converted to `value` by `operator==`.

vinniefalco: That could in theory be quite expensive, can you please open an issue for this in particular?

test/value.cpp
doganulus: Done.

include/boost/json/impl/array.ipp
doganulus: Done.

vinniefalco:
This can't go in until after the current Boost release cycle is over, which will be after 1.76.0 ships in the first week of April.

doganulus:
What about this feature? Could be merged now?

grisumbras:
Please, rebase on develop and push again. Meanwhile by the end of the week I'll review the PR.

grisumbras (approved):

grisumbras:
Thanks for your contribution!

pdimov:
Might be a good idea to add `hash_value` overloads while we're at it (so that `boost::hash` works.) These can just return the result of `std::hash`, or vice versa.

vinniefalco:
> Might be a good idea to add hash_value overloads while we're at it

Open a new issue?

PR #530, opened 2021-03-16
Drone CI uses newer OS images

sdarwin:


sdarwin:
This change was merged. Closing.

PR #541, opened 2021-04-14
Use Docca jam module

grisumbras:
This takes advantage of the Jam module that was added to Docca's develop branch.
* Everything that is already done by the Docca's Jam module was removed from `doc/Jamfile`.
* There's no need now to copy `reference.qbk` into source directory, so the corresponding gitignore line was removed.
* Doxygen configuration file is also generated by the Docca's Jam module. Relevant Doxygen parameters are set up via `<doxygen:param>` properties. As a result, `doc/source.dox` was removed, the parameters _that differ from Doxygen defaults_ were put in `doc/Jamfile`.


vinniefalco:
This is pretty nice

vinniefalco:
This works for me locally, so 👍🏻 merge as you wish. We can do the merges to master after the release.

PR #543, opened 2021-04-21
Implicit conversion from json::string to std::string_view

### LINKED ISSUE #538: Add support for json::string to std::string_view conversion
When I'm using Boost.JSON along with Boost.Beast, it's common to parse a json frame from a remote http/websocket server and do some processing with it.

A typical scenario is, I need to pass a `json::string` to a function where a `std::string_view` is expected.

I can use standalone Boost.JSON with macro `BOOST_JSON_STANDALONE` to replace `boost::string_view` with `std::string_view`. But since I also need Boost.Beast, mixing Boost and standalone Boost.JSON would result in a lot of "macro redefinition" error.

So now I have to tediously do the conversion like this everytime

```cpp
void process_sv(std::string_view sv);

auto j_str = j_v.as_string();
auto j_sv = std::string_view(j_str.data(), j_str.size());

process_sv(j_sv);
```

It would make it much more convenient if a direct conversion from `json::string` to `std::string_view` is available.

Thank you!

grisumbras:
Fix #538

### DIFF
--- a/doc/qbk/release_notes.qbk
+++ b/doc/qbk/release_notes.qbk
@@ -13,6 +13,11 @@
 
 [heading Boost 1.77.0]
 
+[*New Features]
+
+* [issue 538] [link json.ref.boost__json__string.operator_std__string_view
+  Implicit conversion operator from `string` to `std::string_view`].
+
 [*Fixes]
 
 * __object__ deallocates the correct size.
--- a/include/boost/json/string.hpp
+++ b/include/boost/json/string.hpp
@@ -1134,6 +1134,26 @@ public:
         return {data(), size()};
     }
 
+#if ! defined(BOOST_JSON_STANDALONE) && \
+    ! defined(BOOST_NO_CXX17_HDR_STRING_VIEW)
+    /** Convert to a `std::string_view` referring to the string.
+
+        Returns a string view to the underlying character string. The size of
+        the view does not include the null terminator.
+
+        This overload is not defined when either `BOOST_JSON_STANDALONE` or
+        `BOOST_NO_CXX17_HDR_STRING_VIEW` is defined.
+
+        @par Complexity
+
+        Constant.
+    */
+    operator std::string_view() const noexcept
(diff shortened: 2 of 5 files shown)


vinniefalco:
Because this is an implicit conversion I'd like @pdimov or @glenfe to weigh in on whether there are situations where a cycle of conversions could cause compilation errors or unexpected results.

pdimov:
It's already present in standalone.

vinniefalco:
Looks good

vinniefalco:
We need a CHANGELOG.md entry

vinniefalco:
"json::string is convertible to string_view" would be my choice of commit message

doc/qbk/release_notes.qbk
```diff
@@ -13,6 +13,11 @@
 
 [heading Boost 1.77.0]
 
+[*New Features]
+
+* [issue 538] [link json.ref.boost__json__string.operator_std__string_view
+  Implicit conversion operator from `string` to `std::string_view`].
```
grisumbras, line 19: This release notes item will be a link to the operator overload in question. Is this OK @vinniefalco ?

include/boost/json/string.hpp
```diff
@@ -1134,6 +1134,26 @@ class string
...
 
+#if ! defined(BOOST_JSON_STANDALONE) && \
+    ! defined(BOOST_NO_CXX17_HDR_STRING_VIEW)
+    /** Convert to a `std::string_view` referring to the string.
+
+        Returns a string view to the underlying character string. The size of
+        the view does not include the null terminator.
+
+        This overload is not defined when either `BOOST_JSON_STANDALONE` or
+        `BOOST_NO_CXX17_HDR_STRING_VIEW` is defined.
```
grisumbras, line 1145: @vinniefalco What do you think about this paragraph?

vinniefalco: its fine

PR #544, opened 2021-04-23
bugfix: object::table::deallocate to be freed bytes size is inverted

KoaLaYT:
When deallocate, the `bytes` to be freed is not consistent with allocation.
Maybe just a typo.

### DIFF
--- a/include/boost/json/impl/object.hpp
+++ b/include/boost/json/impl/object.hpp
@@ -104,7 +104,7 @@ struct alignas(key_value_pair)
     {
         if(p->capacity == 0)
             return;
-        if(p->is_small())
+        if(! p->is_small())
             sp->deallocate(p,
                 sizeof(table) + p->capacity * (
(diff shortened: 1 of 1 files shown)


grisumbras:
That's a great catch! Please, squash the commits, though.

grisumbras:
@vinniefalco if you're OK with that, I can add a test and a changelog entry. Would you prefer all that to be in one commit?

vinniefalco:
How did this get past testing???!!! 

> I can add a test and a changelog entry. Would you prefer all that to be in one commit?

Yes please

vinniefalco:
Please preserve the authorship even if you squash and add the test and CHANGELOG entry

grisumbras:
Certainly

grisumbras:
The fix is merged in a53e218. Thanks again for your contribution @KoaLaYT.

PR #546, opened 2021-04-24
Add new example 'if_contains'

maksymilian-palka:
The motivation is to have a method checking if the nested object exists or not, for example:

`if_contains(&jv, "nested", "array", 1, "array.2.o");`

It will return a const pointer to the value if it exists at a given path or nullptr otherwise.


vinniefalco:
Does the JSON Pointer specification also solve this problem?

maksymilian-palka:
> Does the JSON Pointer specification also solve this problem?

Yes, usage like:
`if_contains(&jv, "/nested/array/1/array.2.o");`
would be also fine.

Would you like me to provide the exemplary implementation of it or is it already available?

maksymilian-palka:
I have checked JSON Pointer specification in more detail and its implementation deserves a separate PR.

Is this PR good to be merged, or is it redundant/superfluous or is there something still missing?

vinniefalco:
Nothing wrong with an example I suppose, @grisumbras ?

grisumbras:
I kinda feel that with JSON Pointer (which I plan to have full support for by the end of May) this example becomes completely unnecessary.

maksymilian-palka:
@grisumbras: you are correct. If JSON Pointer is coming in Boost::Json 1.77, then this example is redundant. Do you need some help with your implementation? Maybe I could write some tests?

@vinniefalco: should I just close this PR, or do you have an option to delete this PR?

vinniefalco:
> do you have an option to delete this PR?

There's no way to delete a pull request or issue that I know of

maksymilian-palka:
In this case, I close the PR.

@grisumbras, please write me on Slack in case you would like to have some help with the implementation or testing.

PR #547, opened 2021-04-25
Bugfix/object table deallocate

grisumbras:
This PR is a derivative of #544 and only exists to run CI with added test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+Boost 1.77.0:
+
+* `object` deallocates the correct size.
+
 Boost 1.76.0:
 
 * Refactored `value_from` implementation; user customizations are now always
--- a/test/object.cpp
+++ b/test/object.cpp
@@ -24,6 +24,39 @@
 
 BOOST_JSON_NS_BEGIN
 
+
+namespace {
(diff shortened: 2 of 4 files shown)


test/object.cpp
```diff
@@ -1469,6 +1502,26 @@ class object_test
...
+    void
+    testAllocation()
+    {
+        counting_resource res;
+        {
+            storage_ptr sp(&res);
+            object o(sp);
+            o.reserve(1);
+        }
+        BOOST_TEST(res.allocated == res.deallocated);
```
grisumbras, line 1514: @glenfe can you comment on whether this is a correct way to test that the same amount which was allocated was then deallocated?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+Boost 1.77.0:
+
+* Fixed memory deallocation bug in `object`.
```
grisumbras, line 3: @vinniefalco are you satisfied with this changelog entry? Also, see release notes entry below.

test/object.cpp
glenfe: @grisumbras Yes, it is.

PR #551, opened 2021-04-28
Stream cleanup

### LINKED ISSUE #213: Too many streams
There are just too many here:

https://github.com/CPPAlliance/json/blob/develop/include/boost/json/detail/stream.hpp

Surely they are not all needed...

grisumbras:
Fix #213

### DIFF
--- a/include/boost/json/basic_parser.hpp
+++ b/include/boost/json/basic_parser.hpp
@@ -88,12 +88,12 @@ BOOST_JSON_NS_BEGIN
     the error code to a suitable value. This error
     code will be returned by the write function to
     the caller.
-\n            
+\n
     Handlers are required to declare the maximum
     limits on various elements. If these limits
     are exceeded during parsing, then parsing
     fails with an error.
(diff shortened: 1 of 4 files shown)


vinniefalco:
These affect performance

grisumbras:
Yeah, I see that. Do you know the reason? The changes should have been purely cosmetical?

vinniefalco:
> The changes should have been purely cosmetical?

The local stream allows the compiler to assume there is no aliasing

vinniefalco:
@sdkrystian knows more

sdkrystian:
@grisumbras @vinniefalco is correct -- creating the local streams allows for the compiler to assume no aliasing. Without them, it must operate under the assumption that the referenced stream may be modified by writes through aliasable pointer types within the function (or ones deeper within the call stack).

vinniefalco:
In retrospect maybe we should have left a comment on the declaration for `local_stream` explaining intent.

PR #561, opened 2021-05-03
Remove unused ryu functions

### LINKED ISSUE #545: thread_local is not supported on some 32bit platforms
### Version of Boost

1.76.0

Single main including `boost/json.hpp`, building for iOS simulator x86 (32-bit)

I do get the following error:

>./boost/json/detail/ryu/impl/d2s.ipp:733:12: error: thread-local storage is not supported for the current target
>   static thread_local char result[25];

grisumbras:
Fix #545


grisumbras:
So, the coverage reduction(as far as I can tell) is due to the decreased number of lines of code.

The benchmark results are kinda strange: the functions that were removed weren't even used.

PR #568, opened 2021-05-07
Do not crash in array constructor from iterators that form an empty range

### LINKED ISSUE #567: Empty iterator pair construction of array fails
This snippet fails
```
std::vector<std::string> v;
auto  jv    = json::value();
auto& obj   = jv.emplace_object();
obj["debs"] = json::array(v.begin(), v.end());
auto msg = json::serialize(jv);
```

The problem is in the constructor which takes a pair of iterators. When the range is empty it should not allocate.

grisumbras:
Fix #567


grisumbras:
What would be different in those tests?

pdimov:
They would test random access and bidirectional iterators.

PR #569, opened 2021-05-07
to_number overload that takes error

### LINKED ISSUE #548: refactor value::to_number
We can add a set of private member functions which return the error code enum instead, e.g.:
```
    template<class T>
    T value::to_number_impl( error& e ) const noexcept ->
        typename std::enable_if<
            std::is_signed<T>::value &&
            ! std::is_floating_point<T>::value,
                T>::type;
```

And then re-implement the public overloads of `to_number` in terms of this.

grisumbras:
Fix #548


vinniefalco:
Commit message could read "to_number overload takes error"

pdimov (approved):
LGTM

PR #572, opened 2021-05-22
key_value_pair allocates with the correct alignment 

grisumbras:


test/object.cpp
```diff
@@ -1512,6 +1512,12 @@ class object_test
             object o(sp);
             o.reserve(1000);
         }
+
+        res = checking_resource(default_resource);
+        {
+            storage_ptr sp(&res);
+            object o({std::make_pair("one", 1)}, sp);
+        }
```
vinniefalco, line 1520: Just the most minor of nits, I would put a scope around this, e.g.:
```
    {
        res = checking_resource(default_resource);
        {
            storage_ptr sp(&res);
            object o({std::make_pair("one", 1)}, sp);
        }
    }
```

test/checking_resource.hpp
```diff
@@ -0,0 +1,63 @@
...
+BOOST_JSON_NS_BEGIN
+
+class checking_resource
+    : public memory_resource
+{
+    memory_resource* base_;
+    std::map<void const*, std::pair<std::size_t, std::size_t>> allocs_;
+
+public:
+    explicit checking_resource(memory_resource& base) noexcept
```
vinniefalco, line 27: The parameter type should be `storage_ptr`. See:
https://github.com/boostorg/json/blob/develop/include/boost/json/monotonic_resource.hpp#L164

test/object.cpp
```diff
@@ -1530,21 +1498,20 @@ class object_test
     void
     testAllocation()
     {
-        counting_resource res;
+        auto& default_resource = *detail::default_resource::get();
```
vinniefalco, line 1501: Tests should try as hard as possible to only use public interfaces

test/object.cpp
```diff
@@ -1530,21 +1498,20 @@ class object_test
...
+        checking_resource res(default_resource);
         {
             storage_ptr sp(&res);
             object o(sp);
             o.reserve(1);
         }
-        BOOST_TEST(res.allocated == res.deallocated);
 
-        res = {};
+        res = checking_resource(default_resource);
```
vinniefalco, line 1509: Each of these scopes should look something like this:
```
    {
        auto sp = make_shared_resource<checking_resource>();
        
        // this should also work
        //auto sp = make_shared_resource<checking_resource>( storage_ptr() );

        object o( sp );
        o.reserve(1);
}
```

vinniefalco: Alternatively, if you want a non-counted resource:
```
Each of these scopes should look something like this:
```
    {
        checking_resource mr;
        
        // this should also work
        //checking_resource mr( storage_ptr() );

        object o( &mr );
        o.reserve(1);
    }
```

File: test/object.cpp [hunk 5]
```diff
@@ -23,42 +23,10 @@
...
 #include "test.hpp"
 #include "test_suite.hpp"
+#include "checking_resource.hpp"
 
 BOOST_JSON_NS_BEGIN
 
-
-namespace {
-
-struct counting_resource
```
vinniefalco, line 32: LOL my old class was weaksauce

PR #574, opened 2021-05-22
Array value_to/value_from conversion

grisumbras:


test/value_from.cpp
```diff
@@ -205,6 +206,18 @@ class value_from_test
...
+            std::array<int, 1000> a;
+            a.fill(0);
+
+            value b;
+            auto& b_arr = b.emplace_array();
+            b_arr.insert(b_arr.end(), a.begin(), a.end());
+
+            value c = value_from(a);
+            BOOST_TEST(c.is_array());
+            BOOST_TEST(serialize(c) == serialize(b));
```
vinniefalco, line 219: I think you can use `BOOST_TEST( c == b )` now (?)

include/boost/json/detail/value_to.hpp
```diff
@@ -34,6 +34,61 @@ T value_to(U const&);
...
+    cont.reserve(size);
+}
+
+template<typename T>
+void
+try_reserve(
+    T&,
+    std::size_t,
+    priority_tag<0>)
+{ }
```
vinniefalco, line 71: Please put the braces on separate lines this helps coverage

PR #575, opened 2021-05-23
JSON pointer implementation

grisumbras:
* Class `pointer` in `boost/json/pointer.hpp`; stores a `string_view` that should contain a JSON Pointer string.
* Member function `value& value::at(pointer)` that either returns a reference to corresponding child or throws `system_error`.
* Member function `value* value::find(pointer, error_code&)` that either returns a pointer to corresponding child or sets `error_code` to some error.
* `const` versions of those functions.
* Extend `error` and `error_condition` enums with values needed for Pointer.

### DIFF
--- a/include/boost/json/detail/object.hpp
+++ b/include/boost/json/detail/object.hpp
@@ -11,10 +11,12 @@
 #define BOOST_JSON_DETAIL_OBJECT_HPP
 
 #include <boost/json/storage_ptr.hpp>
+#include <boost/json/string_view.hpp>
 #include <cstdlib>
 
 BOOST_JSON_NS_BEGIN
 
+class object;
 class value;
 class key_value_pair;
 
@@ -72,6 +74,19 @@ public:
         return data;
     }
 };
+
+template<class CharRange>
+std::pair<key_value_pair*, std::size_t>
+find_in_object(
+    object const& obj,
+    CharRange key) noexcept;
+
+extern template
+BOOST_JSON_DECL
+std::pair<key_value_pair*, std::size_t>
+find_in_object<string_view>(
+    object const&,
+    string_view key) noexcept;
 
 } // detail
 BOOST_JSON_NS_END
--- a/include/boost/json/error.hpp
+++ b/include/boost/json/error.hpp
@@ -76,6 +76,31 @@ enum class error
 
     /// test failure
     test_failure,
+
+    //
+    // JSON Pointer errors
+    //
+
+    /// missing slash character before token reference
+    missing_slash,
+
+    /// invalid escape sequence
+    invalid_escape,
+
+    /// token should be a number but cannot be parsed as such
+    token_not_number,
+
+    /// current value is neither an object nor an array
+    value_is_scalar,
+
+    /// current value does not contain referenced value
+    not_found,
+
+    /// token cannot be represented by std::size_t
+    token_overflow,
+
+    /// past-the-end index is not supported
+    past_the_end,
 };
 
 /** Error conditions corresponding to JSON errors
@@ -86,7 +111,13 @@ enum class condition
     parse_error = 1,
 
     /// An error on assignment to or from a JSON value
-    assign_error
+    assign_error,
+
+    /// An error related to parsing JSON pointer string
+    pointer_parse_error,
+
+    /// An error related to applying JSON pointer string to a value
+    pointer_use_error,
 };
 
 BOOST_JSON_NS_END
--- a/include/boost/json/impl/error.ipp
+++ b/include/boost/json/impl/error.ipp
@@ -49,6 +49,14 @@ case error::not_number: return "not a number";
 case error::not_exact: return "not exact";
 
 case error::test_failure: return "test failure";
+
+case error::missing_slash: return "missing slash character";
+case error::invalid_escape: return "invalid escape sequence";
+case error::token_not_number: return "token is not a number";
+case error::value_is_scalar: return "current value is scalar";
+case error::not_found: return "no referenced value";
+case error::token_overflow: return "token overflow";
+case error::past_the_end: return "past-the-end token not supported";
             }
         }
 
@@ -80,6 +88,17 @@ case error::exception:
 case error::not_number:
 case error::not_exact:
     return condition::assign_error;
+
+case error::missing_slash:
+case error::invalid_escape:
+    return condition::pointer_parse_error;
+
+case error::token_not_number:
+case error::value_is_scalar:
+case error::not_found:
+case error::token_overflow:
+case error::past_the_end:
+    return condition::pointer_use_error;
             }
         }
     };
@@ -112,6 +131,11 @@ make_error_condition(condition c)
                 return "A JSON parse error occurred";
             case condition::assign_error:
                 return "An error occurred during assignment";
+            case condition::pointer_parse_error:
+                return "A JSON Pointer parse error occurred";
+            case condition::pointer_use_error:
+                return "An error occurred when JSON Pointer was used with"
+                    " a value";
             }
         }
     };
--- a/include/boost/json/impl/object.ipp
+++ b/include/boost/json/impl/object.ipp
@@ -23,6 +23,54 @@
 #include <type_traits>
 
 BOOST_JSON_NS_BEGIN
+namespace detail {
+
+template<class CharRange>
+std::pair<key_value_pair*, std::size_t>
+find_in_object(
+    object const& obj,
+    CharRange key) noexcept
+{
+    BOOST_ASSERT(obj.t_->capacity > 0);
+    if(obj.t_->is_small())
+    {
+        auto it = &(*obj.t_)[0];
+        auto const last =
+            &(*obj.t_)[obj.t_->size];
+        for(;it != last; ++it)
+            if( key == it->key() )
+                return { it, 0 };
+        return { nullptr, 0 };
+    }
+    std::pair<
+        key_value_pair*,
+        std::size_t> result;
+    BOOST_ASSERT(obj.t_->salt != 0);
+    result.second = detail::digest(key.begin(), key.end(), obj.t_->salt);
+    auto i = obj.t_->bucket(
+        result.second);
+    while(i != object::null_index_)
+    {
+        auto& v = (*obj.t_)[i];
+        if( key == v.key() )
+        {
+            result.first = &v;
+            return result;
+        }
+        i = access::next(v);
+    }
+    result.first = nullptr;
+    return result;
+}
+
+
+template
+std::pair<key_value_pair*, std::size_t>
+find_in_object<string_view>(
+    object const& obj,
+    string_view key) noexcept;
+
+} // namespace detail
 
 //----------------------------------------------------------
 
@@ -38,7 +86,7 @@ digest(string_view key) const noexcept
 {
     BOOST_ASSERT(salt != 0);
     return detail::digest(
-        key.data(), key.size(), salt);
+        key.begin(), key.end(), salt);
 }
 
 auto
@@ -176,7 +224,7 @@ object(detail::unchecked_object&& uo)
             access::construct_key_value_pair(
                 dest, pilfer(src[0]), pilfer(src[1]));
             src += 2;
-            auto result = find_impl(dest->key());
+            auto result = detail::find_in_object(*this, dest->key());
             if(! result.first)
             {
                 ++dest;
@@ -409,7 +457,7 @@ insert(
         for(auto& iv : init)
         {
             auto result =
-                find_impl(iv.first);
+                detail::find_in_object(*this, iv.first);
             if(result.first)
             {
                 // ignore duplicate
@@ -564,7 +612,7 @@ find(string_view key) noexcept ->
     if(empty())
         return end();
     auto const p =
-        find_impl(key).first;
+        detail::find_in_object(*this, key).first;
     if(p)
         return p;
     return end();
@@ -578,7 +626,7 @@ find(string_view key) const noexcept ->
     if(empty())
         return end();
     auto const p =
-        find_impl(key).first;
+        detail::find_in_object(*this, key).first;
     if(p)
         return p;
     return end();
@@ -591,8 +639,8 @@ contains(
 {
     if(empty())
         return false;
-    return find_impl(
-        key).first != nullptr;
+    return detail::find_in_object(*this, key).first
+        != nullptr;
 }
 
 value const*
@@ -625,45 +673,6 @@ if_contains(
 
 auto
 object::
-find_impl(
-    string_view key) const noexcept ->
-        std::pair<
-            key_value_pair*,
-            std::size_t>
-{
-    BOOST_ASSERT(t_->capacity > 0);
-    if(t_->is_small())
-    {
-        auto it = &(*t_)[0];
-        auto const last =
-            &(*t_)[t_->size];
-        for(;it != last; ++it)
-            if(key == it->key())
-                return { it, 0 };
-        return { nullptr, 0 };
-    }
-    std::pair<
-        key_value_pair*,
-        std::size_t> result;
-    result.second = t_->digest(key);
-    auto i = t_->bucket(
-        result.second);
-    while(i != null_index_)
-    {
-        auto& v = (*t_)[i];
-        if(v.key() == key)
-        {
-            result.first = &v;
-            return result;
-        }
-        i = access::next(v);
-    }
-    result.first = nullptr;
-    return result;
-}
-
-auto
-object::
 insert_impl(
     pilfered<key_value_pair> p) ->
         std::pair<iterator, bool>
@@ -672,7 +681,7 @@ insert_impl(
     // for preventing aliasing.
     reserve(size() + 1);
     auto const result =
-        find_impl(p.get().key());
+        detail::find_in_object(*this, p.get().key());
     if(result.first)
         return { result.first, false };
     return { insert_impl(
@@ -836,7 +845,7 @@ std::hash<::boost::json::object>::operator()(
     std::size_t seed = jo.size();
     for (const auto& kv_pair : jo) {
         auto const hk = ::boost::json::detail::digest(
-            kv_pair.key().data(), kv_pair.key().size(), 0);
+            kv_pair.key().begin(), kv_pair.key().end(), 0);
         auto const hkv = ::boost::json::detail::hash_combine(
             hk,
             std::hash<::boost::json::value>{}(kv_pair.value()));
--- a/include/boost/json/impl/pointer.ipp
+++ b/include/boost/json/impl/pointer.ipp
@@ -1,0 +1,352 @@
+//
+// Copyright (c) 2022 Dmitry Arkhipov (grisumbras@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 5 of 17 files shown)


vinniefalco:
Maybe we should remove all the trailing whitespace in the entire repo _once_ instead of doing it in every pull request?

include/boost/json/error.hpp
```diff
@@ -213,6 +213,28 @@ enum class error
...
+    token_too_large,
+
+    /// token should be a number but cannot be parsed as such
+    token_not_number,
+
+    /// current value is neither an object nor an array
+    value_is_scalar,
+
+    /// current value does not contain referenced value
+    not_found,
```
vinniefalco, line 237: trailing comma on the last element, is this valid C++?

include/boost/json/impl/error.ipp
```diff
@@ -110,6 +127,11 @@ make_error_condition(condition c)
                 return "A JSON parse error occurred";
             case condition::assign_error:
                 return "An error occurred during assignment";
+            case condition::pointer_parse_error:
+                return "A JSON pointer parse error occurred";
```
vinniefalco, line 131: "JSON Pointer" note capitalization

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,203 @@
...
+            break;
+        }
+
+        if ( c < '0' || c > '9' )
+        {
+            ec = error::token_not_number;
+            return {};
+        }
+
+        result = result * 10 + c - '0';
```
vinniefalco, line 47: You have to check for overflow here

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,203 @@
...
+    {
+        if ( *cur++ != '/' )
+        {
+            ec = error::missing_slash;
+            return root;
+        }
+
+        if ( auto object = result->if_object() )
+        {
+            char buffer[256];
```
vinniefalco, line 126: Hard coded buffer size? Is overflow possible?

include/boost/json/pointer.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#ifndef BOOST_JSON_POINTER_HPP
+#define BOOST_JSON_POINTER_HPP
+
+#include <boost/json/detail/config.hpp>
+#include <boost/json/value.hpp>
+
+BOOST_JSON_NS_BEGIN
+
+/** Path to JSON value nested element
```
vinniefalco, line 18: These javadocs need to be complete

include/boost/json/pointer.hpp
```diff
@@ -0,0 +1,45 @@
...
+ */
+class pointer {
```
vinniefalco, line 20: Can we please put opening braces on their own line

include/boost/json/error.hpp
grisumbras: Yes https://eel.is/c++draft/dcl.enum#nt:enum-specifier

include/boost/json/impl/pointer.ipp
grisumbras: I'm going to control buffer size with a macro.

include/boost/json/pointer.hpp
grisumbras: Sure, this is just a draft.

include/boost/json/error.hpp
vinniefalco: Make sure Doxygen doesn't choke on it

include/boost/json/impl/pointer.ipp
vinniefalco: Can't we do better than that?

grisumbras: Using a local buffer allows to make this function `noexcept`.  Also, if we allocate dynamically, we'd have to provide some way for allocation customization and that would clutter the API.

vinniefalco: >  if we allocate dynamically, we'd have to provide some way for allocation customization and that would clutter the API

I don't necessarily agree. We could have a local buffer, and go to the heap if the buffer is insufficient. We don't have to allow customization of the allocation if the memory is allocated and destroyed in the same function (i.e. if it is temporary). It becomes an implementation detail. I have done this in other places, such as the `stream_parser` on a suspend.

include/boost/json/impl/pointer.ipp
```diff
@@ -84,41 +181,46 @@ parse_string_token(
...
         }
-        else
-        {
-            *out++ = c;
-        }
     }
 
-parse_done:
-    return string_view(start, out - start);
+    return {start, b};;
```
vinniefalco, line 206: double semicolon

include/boost/json/impl/error.ipp
```diff
@@ -49,6 +49,13 @@ case error::not_number: return "not a number";
 case error::not_exact: return "not exact";
 
 case error::test_failure: return "test failure";
+
+            case error::missing_slash: return "missing slash character";
+            case error::invalid_escape: return "invalid escape sequence";
+            case error::token_not_number: return "token is not a number";
+            case error::value_is_scalar: return "current value is scalar";
```
vinniefalco, line 56: Why are these indented differently from the others

grisumbras: I actually thought that the old indent is a mistake. But if it's deliberate, I'll move new lines to the left.

include/boost/json/detail/object.hpp
```diff
@@ -73,6 +75,19 @@ class unchecked_object
...
+find_in_object(
+    object const& obj,
+    CharRange key) noexcept;
+
+extern template
+BOOST_JSON_DECL
+std::pair<key_value_pair*, std::size_t>
+find_in_object<string_view>(
+    object const&,
+    string_view key) noexcept;
```
vinniefalco, line 89: The "extern template" thing is a tiny bit risky, in the sense that there is a higher chance it will fail on some compiler.  There is also the question of linkage and export, for shared libraries. It would be better if this could be expressed without explicit function template instantiation, even if it means forward declaring a type specific to JSON Pointer.

include/boost/json/error.hpp
```diff
@@ -213,6 +213,28 @@ enum class error
 
     /// test failure
     test_failure,
+
+    //
+    // pointer errors
```
vinniefalco, line 218: "JSON Pointer errors"

include/boost/json/impl/error.ipp
vinniefalco: Not a mistake. I try to keep code narrow so it can fit in thinner places like smartphones or in an IDE window with 3 file panes. Since this code is declarative rather than imperative (i.e. it simply provides data in the form of strings and not an algorithm in the form of code), I prioritize shrinking the width measured from the first column over scope-correct indentation.

include/boost/json/impl/object.ipp
```diff
@@ -23,6 +23,54 @@
...
+    result.first = nullptr;
+    return result;
+}
+
+
+template
+std::pair<key_value_pair*, std::size_t>
+find_in_object<string_view>(
+    object const& obj,
+    string_view key) noexcept;
```
vinniefalco, line 71: What is this

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+namespace detail {
+
+
+class pointer_token
+{
+public:
+    class iterator;
+
+    pointer_token(
+        string_view::const_iterator b,
```
vinniefalco, line 26: kind of a long winded way to write `char const*` no?

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+        case '~':
+            c = base_[1];
+            if ( '0' == c )
+            {
+                return '~';
+            }
+            else
+            {
+                BOOST_ASSERT('1' == c);
+                return '/';
```
vinniefalco, line 61: Style: I would write this as
```
if ( '0' == c )
    return '~';
BOOST_ASSERT('1' == c);
return '/';
```

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+    iterator& operator++() noexcept
+    {
+        if ( '~' == *base_ )
+        {
+            base_ += 2;
+        }
+        else
+        {
+            ++base_;
+        }
```
vinniefalco, line 77: I prefer not to use the braces for single statements, e.g.
```
        if ( '~' == *base_ )
            base_ += 2;
        else
            ++base_;
        ```

File: include/boost/json/impl/pointer.ipp [hunk 15]
```diff
@@ -0,0 +1,329 @@
...
+    auto const t_e = token.end();
+    auto s_b = sv.begin();
+    auto const s_e = sv.end();
+    for (; t_b != t_e && s_b != s_e; ++t_b, ++s_b)
+    {
+        if (*t_b != *s_b)
+        {
+            return false;
+        }
+    }
```
vinniefalco, line 122: Too many statements on the same line, which confuses coverage and makes setting breakpoints more difficult. I would rewrite it to use fewer compound expressions:
```
bool
operator==(
    pointer_token token,
    string_view sv) noexcept
{
    auto n = sv.size();
    auto s = sv.data();
    auto p = token.begin();
    auto const end = token.end();
    while(n--)
    {
        if(p == end)
            return false;
        if(*p != *s)
            return false;
        ++p;
        ++s;
    }
    if(p != end)
        return false;
    return true;
}
```

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+    return t_e == t_b && s_e == s_b;
+}
+
+std::size_t
+parse_number_token(
+    string_view::const_iterator& b,
+    string_view::const_iterator e,
+    error_code& ec) noexcept
+{
+    if ( ( b == e ) || ( *b == '0' && ( b + 1 ) != e  && b[1] != '/' ) )
```
vinniefalco, line 133: LOL... coverage! Put each condition on its own line...
```
if (
    ( b == e ) ||
    ( *b == '0' && ( b + 1 )  != e && b[1] != '/' ) )
```

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+
+    std::size_t result = 0;
+    for (; b != e; ++b)
+    {
+        char const c = *b;
+
+        if ( '/' == c)
+        {
+            break;
+        }
```
vinniefalco, line 147: No braces needed

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+
+        if ( c < '0' || c > '9' )
```
vinniefalco, line 149: Would `if( static_cast<unsigned char>(c - '0') > 9 )` be better?

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+        constexpr std::size_t max_div10 = max_size / 10;
+
+        if ( result <= max_div10 )
+        {
+            result *= 10;
+            std::size_t d = c - '0';
+            if ( result <= max_size - d )
+            {
+                result += d;
+                continue;
```
vinniefalco, line 166: There's an easier way to check for unsigned overflow... just test if the new value is _smaller_ than the old value (i.e. it wrapped around).

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+    return pointer_token(start, b);
+}
+
+value*
+if_contains_token(object const& obj, pointer_token token)
+{
+    if(obj.empty())
+    {
+        return nullptr;
+    }
```
vinniefalco, line 222: braces

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+
+    auto const it = detail::find_in_object(obj, token).first;
+    if(!it)
+    {
+        return nullptr;
+    }
```
vinniefalco, line 228: braces

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+    auto cur = ptr_str.begin();
+    while ( ptr_str.end() != cur )
+    {
+        if ( *cur++ != '/' )
+        {
+            ec = error::missing_slash;
+            return nullptr;
+        }
+
+        if ( auto object = result->if_object() )
```
vinniefalco, line 249: sus variable name... e.g. use `po` instead.

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+            auto const token = get_token(cur, ptr_str.end(), ec);
+            if ( ec )
+            {
+                return nullptr;
+            }
+
+            result = detail::if_contains_token(*object, token);
+            cur = token.end().base();
+        }
+        else if ( auto array = result->if_array() )
```
vinniefalco, line 260: sus variable name ... "`pa`" maybe?

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+        {
+            std::size_t index = detail::parse_number_token(cur, ptr_str.end(),
```
vinniefalco, line 262: I would use `auto` here

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+                    ec);
+            if ( ec )
+            {
+                return nullptr;
+            }
```
vinniefalco, line 267: braces

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+
+            result = array->if_contains(index);
```
vinniefalco, line 269: `array` names a type... red flags should be going off in your head when you see this line

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+template <class Value>
+Value&
+at_pointer(Value& root, pointer ptr)
+{
+    error_code ec;
+    auto const found = find_pointer(root, ptr, ec);
+    if ( !found )
+    {
+        detail::throw_system_error(ec, BOOST_JSON_SOURCE_POS);
+    }
```
vinniefalco, line 297: braces

include/boost/json/impl/pointer.ipp
```diff
@@ -0,0 +1,329 @@
...
+    return *found;
+}
+
+} // namespace detail
+
+value const&
+value::at(pointer ptr) const
```
vinniefalco, line 304: You put member function definitions for `value` in pointer.ipp?

include/boost/json/pointer.hpp
```diff
@@ -0,0 +1,46 @@
...
+class pointer
+{
+    string_view str_;
+
+public:
+    /** Default constructor
+    */
+    pointer() noexcept
+        : pointer("")
+    {}
```
vinniefalco, line 29: braces on separate lines for breakpoints and coverage

include/boost/json/pointer.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    /** String constructor
+    */
+    explicit pointer(string_view s) noexcept
+        : str_(s)
+    {}
+
+    string_view
+    string() const noexcept
```
vinniefalco, line 38: sus member function name, it names the type `json::string`

test/pointer.cpp
```diff
@@ -0,0 +1,195 @@
...
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/json
+//
+
+// Test that header file is self-contained.
+#include <boost/json/pointer.hpp>
+// Test header guards
+#include <boost/json/pointer.hpp>
```
vinniefalco, line 13: lol...

include/boost/json/impl/object.ipp
grisumbras: Explicit template instantiation. This puts the template instantiation into the compiled library.

include/boost/json/impl/pointer.ipp
grisumbras: Apparently `string_view::const_iterator` is not `char const*` on msvs 14. I guess, I could try changing it from the other side and use `string_view::const_iterator` less.

include/boost/json/impl/pointer.ipp
grisumbras: Not like that. It would be equivalent to `((int)c - (int)'0') > 9`. But I see how to make something similar work.

include/boost/json/impl/pointer.ipp
grisumbras: I thought it would be reasonable to put all code related to Pointer into a separate file. But if you think it belongs in `value.ipp`, then I'll move it.

include/boost/json/pointer.hpp
grisumbras: Any suggestions for a better name?

include/boost/json/impl/pointer.ipp
vinniefalco: What is the type on msvs 14.1?

include/boost/json/impl/pointer.ipp
```diff
@@ -25,7 +24,8 @@ public:
     pointer_token(
         string_view::const_iterator b,
         string_view::const_iterator e) noexcept
-        : b_(b), e_(e)
+        : b_(b)
+        , e_(e)
```
vinniefalco, line 28: Note that this isn't just style... putting each statement on its own line makes it easier to get a breakpoint, and see coverage.

test/error.cpp
```diff
@@ -65,6 +65,15 @@ class error_test
...
         check(condition::assign_error, error::not_exact);
 
+        check(condition::pointer_parse_error, error::missing_slash);
+        check(condition::pointer_parse_error, error::invalid_escape);
+
+        check(condition::pointer_use_error, error::token_not_number);
+        check(condition::pointer_use_error, error::value_is_scalar);
+        check(condition::pointer_use_error, error::not_found);
+        check(condition::pointer_use_error, error::token_overflow);
+        check(condition::pointer_use_error, error::past_the_end);
```
vinniefalco, line 75: do we need this condition? what would a user do with it?

grisumbras: There are two reasons finding via pointer can fail: invalid pointer string and referencing an element that doesn't exist. This condition is for the latter.

include/boost/json/impl/object.ipp
```diff
@@ -23,6 +23,54 @@
 #include <type_traits>
 
 BOOST_JSON_NS_BEGIN
+namespace detail {
+
+template<class CharRange>
```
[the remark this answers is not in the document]
grisumbras: A range (e.g. `string_view`) whose `value_type` is `char`. I need this as a template for non-allocating implementation of pointer token look up.

PR #580, opened 2021-06-18
Refactor function value::equal, extract int64,uint64,double's equal to a sub function value::number_equal. Readability might be better.

ninghejun:
Refactor function value::equal, extract int64,uint64,double's equal to a sub function value::number_equal. Readability might be better.

### DIFF
--- a/include/boost/json/impl/value.ipp
+++ b/include/boost/json/impl/value.ipp
@@ -389,43 +389,31 @@ destroy() noexcept
 
 bool
 value::
-equal(value const& other) const noexcept
-{
-    switch(kind())
+number_equal(const value &other) const noexcept {
+    switch (kind())
     {
-    default: // unreachable()?
-    case json::kind::null:
-        return other.kind() == json::kind::null;
-
-    case json::kind::bool_:
-        return
-            other.kind() == json::kind::bool_ &&
-            get_bool() == other.get_bool();
-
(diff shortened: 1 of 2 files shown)


vinniefalco:
What problem is this trying to solve? It seems we are adding a new public function, `number_equal`, which has undefined behavior for some inputs.

ninghejun:
> What problem is this trying to solve? It seems we are adding a new public function, `number_equal`, which has undefined behavior for some inputs.

Switch nesting in the equal function is not readable enough. This submission is an attempt to improve readability.

`number_equal` is private, only called by `equal`.  the inputs have been checked by `is_number()` in `equal`.
If we call `number_equal` separately, we get undefined behavior.  It was my fault. I can add parameter checks to the function `number_equal` later.

Is this submission worth it?

vinniefalco:
hmm its pretty readable as it is. There are other considerations here such as code coverage, undefined behavior, And It seems that this change fails tests. Given that the code works and is already pretty understandable, I'm not sure changing it is an improvement. But thank you for taking a look.

ninghejun:
Yes.

I saw 'Enumeration values not handled in switch' in the test result.

These test case failures can be fixed, but result in poorer branch coverage.

Then this change is meaningless.

So this pull request can be closed.

Issue #576, opened 2021-06-02
Test failures on (mostly) macOS


pdimov:
https://travis-ci.org/github/boostorg/json/builds/773287795

s390x fails many tests for non-obvious reasons. One thing to note is that it's a big endian platform (the only one such on Travis.)

Xcode 7.3 and 8.3 fail with:
```
clang-darwin.compile.c++ bin.v2/libs/json/build/clang-darwin-7.3/debug/address-model-32/cxxstd-11-iso/threading-multi/visibility-hidden/src.o
In file included from libs/json/src/src.cpp:10:
In file included from ./boost/json/src.hpp:27:
In file included from ./boost/json.hpp:15:
In file included from ./boost/json/array.hpp:16:
In file included from ./boost/json/storage_ptr.hpp:16:
./boost/json/detail/default_resource.hpp:84:5: error: constexpr constructor never produces a constant expression [-Winvalid-constexpr]
    holder()
    ^
./boost/json/detail/default_resource.hpp:85:11: note: non-literal type 'boost::json::detail::default_resource' cannot be used in a constant expression
        : mr()
          ^
1 error generated.
    "clang++" -x c++ -std=c++11 -fvisibility-inlines-hidden -fPIC -m32 -O0 -fno-inline -Wall -g -fvisibility=hidden -march=i686  -DBOOST_ALL_NO_LIB=1 -DBOOST_CONTAINER_DYN_LINK=1 -DBOOST_JSON_DYN_LINK=1 -DBOOST_JSON_SOURCE -I"."  -c -o "bin.v2/libs/json/build/clang-darwin-7.3/debug/address-model-32/cxxstd-11-iso/threading-multi/visibility-hidden/src.o" "libs/json/src/src.cpp"
...failed clang-darwin.compile.c++ bin.v2/libs/json/build/clang-darwin-7.3/debug/address-model-32/cxxstd-11-iso/threading-multi/visibility-hidden/src.o...
```
Xcode 9.4 fails with
```
clang-darwin.link bin.v2/libs/json/test/doc_quick_look.test/clang-darwin-9.1/debug/cxxstd-11-iso/threading-multi/visibility-hidden/doc_quick_look
Undefined symbols for architecture x86_64:
  "std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::basic_string(char const*)", referenced from:
      boost::json::set3() in doc_quick_look.o
ld: symbol(s) not found for architecture x86_64
[...]

PR #593, opened 2021-07-06
Fix tests failing for 32-bit targets

grisumbras:
When an integer is converted to a floating point type which cannot hold it without loss of precision, the conversion is implementation-defined. This apparently means, that the conversion can result in slightly different results for the same number in different functions. Because of this several tests were failing for some 32 bit targets with optimizations enabled (#576). This change removes those tests and instead adds an equivalent test that uses an integer value that can be converted to float without loss of precision.

The question is, should we remove other similar tests that involve lossy conversions? Also, maybe instead of removing those tests, we should check that the results are within and allowed interval?

### DIFF
--- a/test/value.cpp
+++ b/test/value.cpp
@@ -1473,8 +1473,9 @@ public:
             THRO(std::uint16_t);
             EQUS(std::uint32_t);
             EQUS(std::uint64_t);
-            EQUF(float);
             EQAL(double);
+            // EQUF(float) will result in loss of precision, so the test might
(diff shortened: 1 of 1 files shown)


vinniefalco:
Did this get resolved? Or merged?

grisumbras:
No, I'm working on an alternative that doesn't simply drop `EQUF` tests, but replaces them with different tests.

grisumbras:
#1190 has a 32 bit job that succeeds. So, somehow we appear to have fixed this problem along the way.

PR #606, opened 2021-08-08
Digest calculated in batches of sizeof(size_t)

grisumbras:


vinniefalco:
What is the status of this?

grisumbras:
I've made 6 variants of this and benched them. Here are the results:
https://docs.google.com/spreadsheets/d/1WEfTLx5vEb5TduzzHt-OkAj6HnS4ku4Y4Y7C3YGIgc8/edit?usp=sharing
The 6th approach seems most promising, fairly small regression for a few tests and huge improvements for many tests. But my local benchmarking disagrees with CI as you can see on the picture posted above by the bot.

grisumbras:
Abandoned in favour of #1094.

PR #609, opened 2021-08-17
Fix integer overflow #608

### LINKED ISSUE #608: integer overflow in basic_parser_impl.hpp
Sorry for this lousy report, just wanted to document what I found so far before having to stop for the day. 

oss fuzz found this integer overflow, https://oss-fuzz.com/testcase-detail/4763163676508160

The input to parse is 9e2147483609

One gets an integer overflow: 2147483600 + 57 cannot be represented in type 'int'

57 is the ascii value for '9'

I think I found the issue, it is https://github.com/boostorg/json/blob/develop/include/boost/json/basic_parser_impl.hpp#L2366 carrying out the "+ c" before subtracting '0'.

I believe this might fix it:
```diff
diff --git a/include/boost/json/basic_parser_impl.hpp b/include/boost/json/basic_parser_impl.hpp
index dc926c9..29ba52e 100644
--- a/include/boost/json/basic_parser_impl.hpp
+++ b/include/boost/json/basic_parser_impl.hpp
@@ -2363,7 +2363,7 @@ do_exp3:
                     num.exp == 214748364 && c > '7')))
                 return fail(cs.begin(), error::exponent_overflow);
             ++cs;
-            num.exp = 10 * num.exp + c - '0';
+            num.exp = 10 * num.exp + ( c - '0' );
             continue;
         }
         goto finish_dub;
```

pauldreik:
I added the offending test case to the known crashing inputs so it is tested in CI by the existing fuzz job.

I verified the code change solved the problem.

I also fixed a few more places in the same file which had the same pattern.

fixes #608


grisumbras:
Merged via 3f48a274589aac06280b2c1b0e3d70796cac4b05

Issue #612, opened 2021-08-18
overflow reported by oss fuzz 37341


pauldreik:
the overflow happens at https://github.com/boostorg/json/blob/3f48a274589aac06280b2c1b0e3d70796cac4b05/include/boost/json/basic_parser_impl.hpp#L2397

the error is 

boost/json/basic_parser_impl.hpp:2397:18: runtime error: signed integer overflow: -11 + -2147483640 cannot be represented in type 'int'

PR #613, opened 2021-08-18
add testcase from https://oss-fuzz.com/testcase-detail/6157345913700352

pauldreik:
this adds the testcase for reproducing #612 

to reproduce it locally, run fuzzing/fuzz.sh

once the fuzzers are built, you can run it directly through 
```sh
fuzzing/fuzzer_parser fuzzing/old_crashes/parser/clusterfuzz-testcase-minimized-fuzz_parser-6157345913700352
```


pauldreik:
this is obsoleted by #614 which adds not only this test case but also another.

PR #614, opened 2021-08-18
Fix signed integer overflow in exponent parsing

### LINKED ISSUE #612: overflow reported by oss fuzz 37341

grisumbras:
Fix #612


pauldreik:
I can (in addition to the fuzz CI job with the test cases) confirm this fixes the problem the fuzzer found. I tested all the test cases from https://oss-fuzz.com/testcases?q=group%3A5898123298340864

PR #634, opened 2021-10-05
Parse into types directly

grisumbras:


madmongo1:
@grisumbras bump!

This is a much-missed feature.

grisumbras:
I went with #924 instead.

PR #638, opened 2021-10-10
iostream parsing

### LINKED ISSUE #619: No simple way to parse from std::istream
There should be a quick and convenient way to parse JSON from `std::istream`. E.g. 
```c++
std::ifstream f("path/to/file");
auto jv = json::parse(f);
```

grisumbras:
Fix #619

### DIFF
--- a/include/boost/json.hpp
+++ b/include/boost/json.hpp
@@ -16,6 +16,7 @@
 #include <boost/json/basic_parser.hpp>
 #include <boost/json/error.hpp>
 #include <boost/json/fwd.hpp>
+#include <boost/json/iostreams.hpp>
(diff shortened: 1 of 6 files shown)


vinniefalco:
You didn't give my man a chance to do it?

pdimov:
I don't get it, why does this need its own streambuf?

grisumbras:
> I don't get it, why does this need its own streambuf?

It doesn't. Vinnie previously directed me at using an output streambuf and I've made one. Then using a streambuf became the simplest way to implement parsing from an istream. Now, the question is whether there is a use case for such a streambuf by itself.

PR #651, opened 2021-11-14
Remove standalone

### LINKED ISSUE #650: Remove standalone mode
Standalone mode makes it hard to take advantage of very useful Boost components.

grisumbras:
Fix #650


include/boost/json/detail/align.hpp
```diff
@@ -64,11 +58,6 @@ struct max_align_t
     std::nullptr_t w;
 };
 
-#else
-using std::align;
-using max_align_t = std::max_align_t;
-#endif
```
phprus, line 70: Boost.JSON requires C++11.
Why do you prefer the Boost instead of the features of the C++11 standard library?

include/boost/json/system_error.hpp
```diff
@@ -42,16 +36,6 @@ generic_category();
...
 
-#else
-
-using error_code = std::error_code;
-using error_category = std::error_category;
-using error_condition = std::error_condition;
-using system_error = std::system_error;
-using std::generic_category;
-
-#endif
```
phprus, line 53: Boost.JSON requires C++11.
Why do you prefer the Boost instead of the features of the C++11 standard library?

grisumbras: As you can see, I simply removed the preprocessor logic for standalone mode. I presume, Boost.Align was used because some of the supported platforms do not have implementation for std::align and std::max_align_t. That is, some of the supported platforms implement only a subset of C++11.

PR #652, opened 2021-11-20
Refactor ci

grisumbras:


vinniefalco:
URL, http_proto, and http_io will need these changes eventually

vinniefalco:
squash it?

example/file.hpp
```diff
@@ -72,7 +72,14 @@ class file
         boost::json::error_code& ec)
     {
         close();
+#if defined(_MSC_VER)
+# pragma warning( push )
+# pragma warning( disable : 4996 )
+#endif
```
vinniefalco, line 78: for clarity of exposition wouldn't it be better to wrap the whole file in this pragma instead of just the one line?

include/boost/json/storage_ptr.hpp
```diff
@@ -458,6 +458,13 @@ class storage_ptr
     make_shared_resource(Args&&... args);
 };
 
+#if defined(_MSC_VER)
+# pragma warning( push )
+# if !defined(__clang__) && _MSC_VER <= 1900
+#  pragma message( "MSVC " )
```
vinniefalco, line 464: Why the message?

PR #655, opened 2021-12-02
Refresh azure-pipelines.yml

sdarwin:
A newer copy of azure-pipelines.yml from boostorg/boost-ci. The interesting result from these tests is that the following combinations don't pass (and have been disabled).   That may have been expected.
- clang 3.5
- clang 3.6
- clang 6, libc++, debug 

You could uncomment the jobs, to view more info.


grisumbras:
Merged via #656

PR #657, opened 2021-12-03
NaN and Infinity support

ZFail:
Аdded support for parsing NaN and Infinity.
Serializer already supports this.


grisumbras:
As far as I know, neither NaN nor infinity is a part of JSON format

ZFail:
> As far as I know, neither NaN nor infinity is a part of JSON format

Yes, but for example comments too.
JSON5 is supporting NaN and Infinity.
Maybe add by option?

pdimov:
What do you accept for NaN or infinity?

ZFail:
> What do you accept for NaN or infinity?

JSON
```
{
  "nan": NaN,
  "infinity": Infinity,
  "minus_infinity": -Infinity
}
```
parsed as
```c++
object obj( {
  { "nan",  std::numeric_limits<double>::quiet_NaN() }, 
  { "infinity", std::numeric_limits<double>::infinity() }, 
  { "minus_infinity", -std::numeric_limits<double>::infinity() }
} );
```

grisumbras:
This needs to be disabled by default and enabled with an option. See 
https://www.boost.org/doc/libs/1_77_0/libs/json/doc/html/json/ref/boost__json__parse_options.html

pdimov:
The [JSON5 spec](https://spec.json5.org/#numbers) also allows `+NaN`, `-NaN` and `+Infinity`:

> The IEEE 754 value positive infinity must be the literal characters Infinity and may be prefixed with an optional plus sign.
> The IEEE 754 value negative infinity must be the literal characters -Infinity.
> The IEEE 754 value NaN must be the literal characters NaN and may be prefixed with an optional plus or minus sign.

glebov-andrey:
Hi!
I would like to pick up this PR (@ZFail has added me to his repo).
We need this feature either way, but I'd like to know if there is interest in merging it.

I've already committed optional support for `Infinity`/`NaN` parsing through `parse_options::allow_infinity_and_nan`.
If there is interest, then I'll add more tests.

Another question is regarding leading `+` and `-NaN` support.
As @pdimov mentioned, JSON5 supports these.
But IMO, supporting leading `+` signs specifically for `Infinity` and `NaN` is not a good idea.
In JSON5 this syntax is supported for numbers in general, which is not true for JSON.
If there is interest in supporting leading `+` signs, then that should either be extended to all numbers (with the parse option renamed to `allow_extended_numbers`), or implemented separately with it's own parse option.

`-NaN` is more of a grey area.
It could definitely make sense to support this, given that the sign of a NaN is actually observable using `std::signbit()`.
But in that case the serializing code should also be modified to produce `-NaN` for such values.
Otherwise we loose the ability to roundtrip numbers, which is the main point of this feature.

test/basic_parser.cpp
```diff
@@ -690,14 +692,22 @@ class basic_parser_test
...
+            make_options(false, true, false, false),
+            make_options(true, true, false, false),
+            make_options(false, false, true, true),
+            make_options(true, false, true, true),
+            make_options(false, true, true, true),
+            make_options(true, true, true, true),
+            make_options(false, false, false, true),
+            make_options(true, false, false, true),
+            make_options(false, true, false, true),
+            make_options(true, true, false, true)
```
vinniefalco, line 710: These would look better using 0 and 1 instead of true and false I bet.. lol

test/serializer.cpp
```diff
@@ -427,6 +427,33 @@ class serializer_test
...
+            BOOST_TEST(sr.read(buf) == "Infinity");
+            BOOST_TEST(sr.done());
+        }
+        {
+            value jv = -std::numeric_limits<double>::infinity();
+            serializer sr;
+            sr.reset(&jv);
+            BOOST_TEST(! sr.done());
+            char buf[32];
+            BOOST_TEST(sr.read(buf) == "-Infinity");
```
vinniefalco, line 454: Is this what Javascript emits, `-Infinity` ?

vinniefalco:
This is still not reviewed and merged?

glebov-andrey:
> This is still not reviewed and merged?

Yeah, it was sort of forgotten about for a while.
Since I needed to rebase this onto 1.82.0 anyway, I decided to take over the PR.

I also left a couple TODOs in the code, if someone could have a look at them:
https://github.com/boostorg/json/pull/657/commits/2dbb08f3fccbe8599e204655a68e3f256f4959ba#diff-2cc706ac1836b6457314bc5f7f00837fc38310395e9d1486782a82b7d5b6e583R105
https://github.com/boostorg/json/pull/657/commits/2dbb08f3fccbe8599e204655a68e3f256f4959ba#diff-ac0f59cf63d51320da1fc37682921e31221528ac2e1c8412363fa3f4d2dead5cR769

test/serializer.cpp
glebov-andrey: No, JS emits `null` for infinities and NaNs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#:~:text=The%20numbers%20Infinity%20and%20NaN%2C%20as%20well%20as%20the%20value%20null%2C%20are%20all%20considered%20null

I'm not actually sure whether these tests should be part of this PR though, considering they are not testing the parser, but decided to leave them just in case for now.

test/basic_parser.cpp
glebov-andrey: Given the growing amount of combinations, I'd rather replace it with a loop (see https://github.com/boostorg/json/pull/657/commits/d665b08517a9d9bbc4475003eb4b5cfa2d920579).

vinniefalco: The loop is smaller but it is easier to understand and read when it is written out. And it is easier to set a breakpoint on a particular combination without the loop..

glebov-andrey:
The massive switch/case in `basic_parser::parse_document()` is not the most easily extendable thing.
Here is a template recursion based replacement, if you're interested: https://github.com/ZFail/json/commit/887a3e8223f6935c86e08f58ee35a13ae5117457
I put it in a separate branch in case such an implementation is not acceptable, e.g. due to debug performance.

test/basic_parser.cpp
glebov-andrey: Ok, done

grisumbras:
Why specifically `Infinity` and `NaN`? My cursory research shows that the only implementation that outputs these is Python.

glebov-andrey:
> Why specifically `Infinity` and `NaN`? My cursory research shows that the only implementation that outputs these is Python.

I think the original reasoning is simple - that's what Boost.JSON serializes them as: https://github.com/boostorg/json/blob/ee01580cfde818e31d8ea41c2e4056bcff9e3018/include/boost/json/detail/ryu/detail/common.hpp#L85-L99
Given that JSON is closely related to JS, and JS has `Infinity` and `NaN` as global properties, this actually seems reasonable.
That's also probably the reasoning behind JSON5's spelling.

grisumbras:
While JS has them, it doesn't parse them as numbers. But I guess, this is a common enough approach to pick it.

One question. I'm not sure we actually need the template parameter for whether special number support is enabled. Can you test if the template parameter affects performance and how much it affects binary size?

Also, please rebase on the current develop and squash. after that I will approve CI runs.

grisumbras:
Are you still planning to do this? I've made a few changes to parsing of JSON literals that makes implementing this change much easier, and also simplifies getting full coverage.

PR #671, opened 2022-01-15
Naturally grow string in serialize_impl

### LINKED ISSUE #668: incorrect behavior of `serialize_impl`
`serialize_impl` is used to serialize JSON into a `std::string`:

https://github.com/boostorg/json/blob/307d190bcfc0707fc532839b3be78e6e3b800519/include/boost/json/impl/serialize.ipp#L21

The problem is that line 52 is incorrect, I think it should read:
```
s.reserve( s.capacity() + 1 );
s.resize( s.capacity() );
```

otherwise, every 2nd trip through the `for` loop will be a 1-char write.

grisumbras:
Fix #668


vinniefalco:
put "Fix #668" in the 3rd line of the commit msg

grisumbras:
The interesting thing about this change is that I have not been able to observe a definitive performance increase.

vinniefalco:
Well it only happens under certain circumstances, where you end up with a 1-char serialization through the loop every other trip. Which is not going to be noticeable in most cases. But the fix is correct.

PR #675, opened 2022-02-05
[serializer] Added handling for NaN & Inf.

### LINKED ISSUE #397: Handling of infinity and NaN
Specifically, if the user puts an inf or NaN into a value, what happens on serialization? We need to specify what should happen on serialization. And if we output "inf" and "nan" then we need to specify what happens on parsing. And we might need to provide a configuration option for allowing it / disallowing it (on parse).

klemens-morgenstern:
Closes #397

### DIFF
--- a/include/boost/json/detail/impl/format.ipp
+++ b/include/boost/json/detail/impl/format.ipp
@@ -111,10 +111,44 @@ format_int64(
 
 unsigned
 format_double(
-    char* dest, double d) noexcept
+    char* dest, double d, const serialize_options & opt) noexcept
 {
-    return static_cast<int>(
-        ryu::d2s_buffered_n(d, dest));
+    if (std::isnan(d))
+    {
+        std::copy(opt.nan.begin(), opt.nan.end(), dest);
+        return opt.nan.size();
+    }
+    else if (std::isinf(d))
+    {
+        if (d > 0.)
+        {
+            std::copy(opt.pinf.begin(), opt.pinf.end(), dest);
+            return opt.pinf.size();
+        }
+        else
+        {
+            std::copy(opt.ninf.begin(), opt.ninf.end(), dest);
+            return opt.ninf.size();
+        }
+    }
+    else
+        return static_cast<int>(
+            ryu::d2s_buffered_n(d, dest));
+}
+
+unsigned
+format_double(
+    char* dest, double d, error_code & ec) noexcept
+{
+    if (std::isnan(d))
+        ec = make_error_code(error::not_number);
+    else if (std::isinf(d))
+        ec = make_error_code(error::exponent_overflow);
+    else
+        return static_cast<int>(
+            ryu::d2s_buffered_n(d, dest));
+
+    return 0u;
 }
 
 } // detail
--- a/include/boost/json/impl/serialize.ipp
+++ b/include/boost/json/impl/serialize.ipp
@@ -20,13 +20,14 @@ static
 void
 serialize_impl(
     std::string& s,
-    serializer& sr)
+    serializer& sr,
+    const serialize_options & opt)
 {
     // serialize to a small buffer to avoid
     // the first few allocations in std::string
     char buf[BOOST_JSON_STACK_BUFFER_SIZE];
     string_view sv;
-    sv = sr.read(buf);
+    sv = sr.read(buf, opt);
     if(sr.done())
     {
         // fast path
@@ -45,7 +46,8 @@ serialize_impl(
     {
         sv = sr.read(
             &s[0] + len,
-            s.size() - len);
+            s.size() - len, 
+            opt);
         len += sv.size();
         if(sr.done())
             break;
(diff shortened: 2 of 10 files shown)


include/boost/json/detail/impl/format.ipp
```diff
@@ -111,10 +111,44 @@ format_int64(
...
 unsigned
 format_double(
-    char* dest, double d) noexcept
+    char* dest, double d, const serialize_options & opt) noexcept
 {
-    return static_cast<int>(
-        ryu::d2s_buffered_n(d, dest));
+    if (std::isnan(d))
+    {
+        std::copy(opt.nan.begin(), opt.nan.end(), dest);
```
vinniefalco, line 118: did you mean
```
opt.copy( dest, 0);
```
?

include/boost/json/detail/impl/format.ipp
```diff
@@ -111,10 +111,44 @@ format_int64(
...
+        return static_cast<int>(
+            ryu::d2s_buffered_n(d, dest));
+}
+
+unsigned
+format_double(
+    char* dest, double d, error_code & ec) noexcept
+{
+    if (std::isnan(d))
+        ec = make_error_code(error::not_number);
```
vinniefalco, line 144: Did you mean
```
ec = error::not_number;
```
?

include/boost/json/impl/serialize.ipp
```diff
@@ -59,62 +61,173 @@ serialize_impl(
...
+static
+void
+serialize_impl(
+    std::string& s,
+    serializer& sr,
+    error_code & ec)
+{
+  // serialize to a small buffer to avoid
+  // the first few allocations in std::string
+  char buf[BOOST_JSON_STACK_BUFFER_SIZE];
```
vinniefalco, line 135: Why are we duplicating this function?

include/boost/json/impl/serializer.ipp
```diff
@@ -675,44 +680,69 @@ write_value(stream& ss)
...
+            return write_object<StackEmpty>(ss, opt);
         }
     }
 }
 
 string_view
 serializer::
 read_some(
-    char* dest, std::size_t size)
+    char* dest, std::size_t size, const serialize_options & opt)
```
vinniefalco, line 720: Why are we passing the options into every call to `read_some`? Shouldn't we construct with the options instead, and then access the options as a data member instead of passing it as a parameter to every member function?

include/boost/json/detail/impl/format.ipp
klemens-morgenstern: I did not mean that, because I don't see why.

include/boost/json/impl/serializer.ipp
```diff
@@ -784,8 +819,10 @@ serializer::
 reset(string const* p) noexcept
 {
     cs0_ = { p->data(), p->size() };
-    fn0_ = &serializer::write_string<true>;
-    fn1_ = &serializer::write_string<false>;
+    fn0_opt_ = &serializer::write_string<true>;
+    fn1_opt_ = &serializer::write_string<false>;
+    fn0_err_ = &serializer::write_string<true>;
+    fn1_err_ = &serializer::write_string<false>;
```
vinniefalco, line 825: hopefully this nonsense will go away when we make the options a data member

include/boost/json/impl/serialize.ipp
klemens-morgenstern: I thought it would turn out different, i.e. need some early breaks for errors. Was wrong, could be redone as a template.

include/boost/json/serialize_options.hpp
```diff
@@ -0,0 +1,33 @@
...
+{
+  /// The value to use for a positive `inf` double.
+  /// The value will not be quoted in json.
+  string_view pinf = "null";
+  /// The value to use for a negative `inf` double.
+  /// The value will not be quoted in json.
+  string_view ninf = "null";
+  /// The value to use for a NaN double.
+  /// The value will not be quoted in json.
+  string_view nan  = "null";
```
vinniefalco, line 28: The defaults are supposed to be empty strings, i.e. throwing on inf/nan is the default behavior.

include/boost/json/serializer.hpp
```diff
@@ -228,10 +238,90 @@ class serializer
 
         @param size The maximum number of characters
         to write to the memory pointed to by `dest`.
+
+        @param opts The options for serialization.
     */
     BOOST_JSON_DECL
     string_view
-    read(char* dest, std::size_t size);
+    read(char* dest, std::size_t size, const serialize_options & opts = {});
```
vinniefalco, line 246: The options should be for the whole serializer not for each individual call to `read`

test/double.cpp
```diff
@@ -421,11 +421,45 @@ class double_test
         }
     };
 
+    void testInvalid()
+    {
+      const json::serialize_options testOpts = {"printed dollars", "government debt", "usd value"};
```
vinniefalco, line 426: These example strings are non-sterile

include/boost/json/impl/serializer.ipp
klemens-morgenstern: This should make the code faster, since the compiler doesn't need to recheck which it is.

vinniefalco:
Exposition (the .qbk files) needs to be updated explaining serialization error handling and format conversion of nan/inf

include/boost/json/serialize_options.hpp
klemens-morgenstern: Probably the wrong default, I copied Javascript here.

include/boost/json/impl/serializer.ipp
klemens-morgenstern: To be fair, it's probably overdoing it, because this is one unnecessary if -else per double overhead.  I didn't profile either, just copied your model.

include/boost/json/detail/impl/format.ipp
vinniefalco: So we don't have to include `<algorithm>`, which you also forgot

include/boost/json/impl/serialize.ipp
vinniefalco: A template would still result in duplication in the emitted code, we need to avoid that since Boost.JSON needs to be as lean as possible.

grisumbras:
The approach we described in the issue is somewhat different:
1. Serializer options are empty by default (not nulls).
2. If an option is empty and the corresponding special value is encountered — an error is produced.
3. Thus, users should be able to use options _and_ error codes together. If an error_code isn't used, the serializer throws.

Top level APIs for serialzer should probably look like this:
```
serializer::serializer(serializer_options const& = {});

serializer::read(char* dest, std::size_t size, serializer_options const& = {});
serializer::read(char* dest, std::size_t size, error_code& ec, serializer_options const& = {});
serializer::read(char* dest, std::size_t size, std::error_code& ec, serializer_options const& = {}); // for convenience
```

vinniefalco:
> Top level APIs for serialzer should probably look like this:

Well no, the serializer should be constructed with the options, they should not be per-read. Alternatively there can be a `set_options` function. But it should be specified that undefined behavior results if the options are changed in the middle of serialization.

include/boost/json/detail/impl/format.ipp
```diff
@@ -111,10 +111,44 @@ format_int64(
...
+    if (std::isnan(d))
+    {
+        std::copy(opt.nan.begin(), opt.nan.end(), dest);
+        return opt.nan.size();
+    }
+    else if (std::isinf(d))
+    {
+        if (d > 0.)
+        {
+            std::copy(opt.pinf.begin(), opt.pinf.end(), dest);
```
vinniefalco, line 125: This code has to handle the case where there is not enough room in the output to hold the user-defined string

grisumbras:
Superceded by #884

PR #678, opened 2022-02-10
s390x architecture drone jobs

sdarwin:
s390x jobs are running at cloud.ibm.com on their new VPC Infrastructure, Tokyo region. 
Drone doesn't include an IBM autoscaler currently, instances are running as provisioned virtual servers.
The json clang-12 job completes in 8 minutes. The json gcc job takes more than 1 hour.  I've commented out the gcc job for now. You could test/debug in a feature branch or fork.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -14,6 +14,8 @@ windowsglobalimage="cppalliance/dronevs2019"
 
 def main(ctx):
   return [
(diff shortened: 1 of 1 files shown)


grisumbras:
Merged via 407fbb809bec96d78dd2a5e8e336356dba0445a9

PR #686, opened 2022-02-28
Refactoring of conversion functions

grisumbras:
The list of changes:
* take advantage of mp11
* split `container_traits` into several traits and helpers
* split `map_traits` into several traits
* use `value_type` more consistently
* refactor `detail::inserter`
* use `detail::inserter` with map-like types
* refactor conversion of tuple-like to array
* remove conversion from `value_ref` and don't use `value_to/from` when constructing `value` from `value_ref`
* check for round-trip
* user-facing conversion traits
* opt-in null-like conversion
* same requirements for `value_to` and `value_from`
* stricter `is_map_like`
* stricter `is_sequence_like`

### DIFF
--- a/doc/qbk/quickref.xml
+++ b/doc/qbk/quickref.xml
@@ -58,6 +58,7 @@
           <member><link linkend="json.ref.boost__json__is_string_like">is_string_like</link></member>
           <member><link linkend="json.ref.boost__json__is_sequence_like">is_sequence_like</link></member>
           <member><link linkend="json.ref.boost__json__is_map_like">is_map_like</link></member>
+          <member><link linkend="json.ref.boost__json__is_null_like">is_null_like</link></member>
           <member><link linkend="json.ref.boost__json__is_tuple_like">is_tuple_like</link></member>
           <member><link linkend="json.ref.boost__json__value_from_tag">value_from_tag</link></member>
           <member><link linkend="json.ref.boost__json__value_to_tag">value_to_tag</link></member>
--- a/doc/qbk/quickref.xml
+++ b/doc/qbk/quickref.xml
@@ -55,6 +55,10 @@
           <member><link linkend="json.ref.boost__json__has_value_from">has_value_from</link></member>
           <member><link linkend="json.ref.boost__json__has_value_to">has_value_to</link></member>
           <member><link linkend="json.ref.boost__json__is_deallocate_trivial">is_deallocate_trivial</link></member>
+          <member><link linkend="json.ref.boost__json__is_string_like">is_string_like</link></member>
+          <member><link linkend="json.ref.boost__json__is_sequence_like">is_sequence_like</link></member>
+          <member><link linkend="json.ref.boost__json__is_map_like">is_map_like</link></member>
+          <member><link linkend="json.ref.boost__json__is_tuple_like">is_tuple_like</link></member>
           <member><link linkend="json.ref.boost__json__value_from_tag">value_from_tag</link></member>
           <member><link linkend="json.ref.boost__json__value_to_tag">value_to_tag</link></member>
         </simplelist>
--- a/include/boost/json/conversion.hpp
+++ b/include/boost/json/conversion.hpp
@@ -19,14 +19,11 @@
 BOOST_JSON_NS_BEGIN
 namespace detail {
 
-using std::begin;
-using std::end;
-
 template<std::size_t I, class T>
 using tuple_element_t = typename std::tuple_element<I, T>::type;
 
 template<class T>
-using value_type = remove_cvref< decltype(*begin(std::declval<T&>())) >;
+using value_type = remove_cvref< decltype(*std::begin(std::declval<T&>())) >;
 template<class T>
 using mapped_type = tuple_element_t< 1, value_type<T> >;
 
@@ -41,9 +38,15 @@ using key_type = mp11::mp_eval_or<
     T>;
 
 template<class T>
+using iterator_type = decltype(std::begin(std::declval<T&>()));
+template<class T>
 using are_begin_and_end_same = std::is_same<
-    decltype(begin(std::declval<T&>())),
-    decltype(end(std::declval<T&>()))>;
+    iterator_type<T>,
+    decltype(std::end(std::declval<T&>()))>;
+
+template<class T>
+using begin_iterator_category = typename std::iterator_traits<
+    iterator_type<T>>::iterator_category;
 
 template<class T>
 using has_positive_tuple_size = mp11::mp_bool<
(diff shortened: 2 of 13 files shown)


include/boost/json/conversion.hpp
```diff
@@ -0,0 +1,293 @@
...
+#ifndef BOOST_JSON_CONVERSION_HPP
+#define BOOST_JSON_CONVERSION_HPP
+
+#include <boost/json/string_view.hpp>
+#include <boost/mp11/algorithm.hpp>
+#include <boost/mp11/utility.hpp>
+
+#include <utility>
+
+BOOST_JSON_NS_BEGIN
```
vinniefalco, line 19: if we don't have standalone we don't need this namespace macro

include/boost/json/conversion.hpp
```diff
@@ -0,0 +1,293 @@
...
+namespace detail {
```
vinniefalco, line 20: Implementation details should be in json/detail or json/impl not a public header since this is scanned by doxygen

include/boost/json/conversion.hpp
```diff
@@ -0,0 +1,293 @@
...
+#ifndef BOOST_JSON_DOCS
+    : mp11::mp_all<
+        mp11::mp_valid_and_true<detail::are_begin_and_end_same, T>,
+        mp11::mp_valid<detail::begin_iterator_category, T>>
+{ };
+#else
+;
+#endif
+
+/** Determine if `T` can be treated like a 1-to-1 mapping during
```
vinniefalco, line 157: Metafunction briefs should start with "Metafunction"

include/boost/json/conversion.hpp
```diff
@@ -0,0 +1,293 @@
...
+        `std::pair<K, M>`; and
+
+    @li <tt>std::is_string_like<K>::value</tt> is `true`; and
+
+    @li given `v`, a glvalue of type `V`, and `E`, the type denoted by
+        `decltype(t.emplace(v))`,
+        <tt>std::is_tuple_like<E>::value</tt> is `true`;
+
+    then the trait provides the member constant `value`
+    that is equal to `true`. Otherwise, `value` is equal to `false`.<br>
```
vinniefalco, line 175: The alternatives should be "Then this is an alias for `std::true_type`,...  otherwise it is an alias for `std::false_type`"

include/boost/json/conversion.hpp
```diff
@@ -0,0 +1,293 @@
...
+
+    @par Types satisfying the trait
+
+    <a href="https://en.cppreference.com/w/cpp/container/map"><tt>std::map</tt></a>,
+    <a href="https://en.cppreference.com/w/cpp/container/unordered_map"><tt>std::unordered_map</tt></a>.
+
+    @see @ref value_from, @ref value_to
+*/
+template<class T, class Enable = void>
+struct is_map_like
```
vinniefalco, line 206: I usually document metafunctions like this
```
#ifdef BOOST_JSON_DOCS
template<class T>
using has_value_to = __see_below__;
#else
...
```

The result looks like this:
image

include/boost/json/detail/value_to.hpp
```diff
@@ -67,152 +69,177 @@ void
...
-        container_traits<T>::value_type;
     const object& obj = jv.as_object();
     T result;
-    try_reserve(result, obj.size(), priority_tag<2>());
-    for (const auto& val : obj)
-        result.insert(value_type{typename map_traits<T>::
-            pair_key_type(val.key()), value_to<typename
-                map_traits<T>::pair_value_type>(val.value())});
+    detail::try_reserve(result, obj.size(), reserve_implementation<T>());
+    std::transform(obj.begin(), obj.end(),
```
vinniefalco, line 218: Does this mean we are including `<algorithm>` now?

test/value_to.cpp
```diff
@@ -114,12 +126,50 @@ class value_to_test
...
+                decltype(std::back_inserter(v)) >::value);
+        }
+        {
+            std::array<int, 2> arr;
+            detail::try_reserve(
+                arr, 2, detail::reserve_implementation<decltype(arr)>());
+        }
+        {
+            int n;
+            detail::try_reserve(
```
vinniefalco, line 149: You're testing an implementation detail directly, which is not something the user would ever do. Better to test this indirectly

doc/qbk/quickref.xml
```diff
@@ -55,6 +55,11 @@
           <member><link linkend="json.ref.boost__json__has_value_from">has_value_from</link></member>
           <member><link linkend="json.ref.boost__json__has_value_to">has_value_to</link></member>
           <member><link linkend="json.ref.boost__json__is_deallocate_trivial">is_deallocate_trivial</link></member>
+          <member><link linkend="json.ref.boost__json__is_string_like">is_string_like</link></member>
+          <member><link linkend="json.ref.boost__json__is_sequence_like">is_sequence_like</link></member>
+          <member><link linkend="json.ref.boost__json__is_map_like">is_map_like</link></member>
+          <member><link linkend="json.ref.boost__json__is_null_like">is_null_like</link></member>
+          <member><link linkend="json.ref.boost__json__is_tuple_like">is_tuple_like</link></member>
```
vinniefalco, line 62: These are not in alphabetical order, was this intended?

include/boost/json/conversion.hpp
vinniefalco: This stuff should go in `<boost/json/detail/conversion.hpp>`

vinniefalco:
This needs exposition in the docs

include/boost/json/conversion.hpp
```diff
@@ -0,0 +1,293 @@
...
+    template <>
+    struct is_tuple_like<your::tuple> : std::false_type
+    { };
+
+    } // namespace boost
+    } // namespace json
+    @endcode
+
+
+    @par Types satisfying the trait
```
vinniefalco, line 240: these are called models, e.g. `@par Models`

include/boost/json/conversion.hpp
grisumbras: These aren't metafunctions, they are type traits.

include/boost/json/conversion.hpp
grisumbras: They aren't aliases, they are separate types. I copied the approach to documenting these from cppreference.

include/boost/json/conversion.hpp
grisumbras: They are not aliases, they are class templates. This is a deliberate decision to allow user specialisations.

include/boost/json/conversion.hpp
vinniefalco: it isn't a function returning a value?

include/boost/json/conversion.hpp
vinniefalco: Oh, yes if they can be specialized then `using` is not correct

include/boost/json/detail/value_to.hpp
grisumbras: We were already including it

test/value_to.cpp
grisumbras: This is specifically a test for these implementation details. Indirect testing would be much more complicated, and probably a bit brittle, because this helpers are simply for memory allocation optimization purposes.

doc/qbk/quickref.xml
grisumbras: No, it was a mistake

include/boost/json/conversion.hpp
grisumbras: No, because this is not a concept or a type requirement. I've asked around and people say that "satifies trait" is correct.

PR #701, opened 2022-05-07
More explicitly document invalidation rules

sehe:
The reference/iterator invalidation rules can surprise people who expect
associative containers to be node-based, similar to `std::[unordered_]map`.

This calls it out more visibly to avoid surprises.


sehe:
I got the snippet reference wrong after all. Sigh. Also, it looks weird because the callout box is hiding the snippet box. Perhaps it's better to put the callout at the bottom, separate from the snippet?

Will update so that we have something tangible to compare.

sehe:
I think this looks better now.

grisumbras:
Merged via 98d37f93706b50fd057d14bd241a2bfbcfec1527

PR #706, opened 2022-05-23
Update Drone - clang-14 and gcc-12

sdarwin:
Before merging, test locally. "failed updating 4 targets" "error: definition of implicit copy assignment operator for 'const_stream' is deprecated"


grisumbras:
Merged via #707

PR #709, opened 2022-05-26
better subview

### LINKED ISSUE #708: 1.79: Build error with gcc 8 in C++17 mode: ambiguous conversion to string_view
### Version of Boost

Boost 1.79.0 release with Boost.JSON patch from [release notes](https://www.boost.org/users/history/version_1_79_0.html) applied.

Build Boost.JSON:

```
b2 -j16 -q -d2  --layout=system --ignore-site-config --user-config=/build/boost/boost_1_79_0/user-config.jam variant=release debug-symbols=on cxxstd=17-gnu numa=on link=static,shared  --without-python
```

This results in compiler errors like this:

```
gcc.compile.c++ bin.v2/libs/json/build/gcc-8/release/cxxstd-17-gnu/debug-symbols-on/link-static/numa-on/threading-multi/visibility-hidden/src.o

    "g++"   -std=gnu++17 -fvisibility-inlines-hidden -g -O2 -fdebug-prefix-map=/build/boost/boost_1_79_0=. -fstack-protector-strong -Wformat -Werror=format-security -O3 -fno-math-errno -fno-rounding-math -fno-signaling-nans -fno-trapping-math -ftree-vectorize -fgcse-sm -fgcse-las -fgcse-after-reload -ftree-loop-im -ftree-loop-distribution -ftree-loop-if-convert -funswitch-loops -frename-registers -fivopts -fvisibility-inlines-hidden -fno-plt -faligned-new -march=nehalem -mtune=skylake -Wno-unused-local-typedefs -Wno-register -Wno-deprecated-declarations -Wno-register -Wno-deprecated-declarations -pthread -O3 -finline-functions -Wno-inline -Wall -g -fvisibility=hidden -Wdate-time -D_FORTIFY_SOURCE=2  -DBOOST_ALL_NO_LIB=1 -DBOOST_CONTAINER_STATIC_LINK=1 -DBOOST_JSON_SOURCE -DBOOST_JSON_STATIC_LINK=1 -DBOOST_USE_NUMA -DNDEBUG  -I"."  -c -o "bin.v2/libs/json/build/gcc-8/release/cxxstd-17-gnu/debug-symbols-on/link-static/numa-on/threading-multi/visibility-hidden/src.o" "libs/json/src/src.cpp"

In file included from ./boost/json/value.hpp:20,
                 from ./boost/json/array.hpp:1756,
                 from ./boost/json.hpp:15,
                 from ./boost/json/src.hpp:27,
                 from libs/json/src/src.cpp:10:
./boost/json/string.hpp: In member function 'int boost::json::string::compare(boost::json::string_view) const':
./boost/json/string.hpp:1948:33: error: call of overloaded 'basic_string_view(const boost::json::string&)' is ambiguous
         return string_view(*this).compare(sv);
                                 ^
In file included from ./boost/json/string_view.hpp:14,
                 from ./boost/json/kind.hpp:14,
                 from ./boost/json/array.hpp:14,
                 from ./boost/json.hpp:15,
                 from ./boost/json/src.hpp:27,
                 from libs/json/src/src.cpp:10:
./boost/core/detail/string_view.hpp:391:5: note: candidate: 'boost::core::basic_string_view<Ch>::basic_string_view(const std::basic_string_view<Ch, std::char_traits<_CharT> >&) [with Ch = char]'
     basic_string_view( std::basic_string_view<Ch, std::char_traits<Ch> > const& str ) BOOST_NOEXCEPT: p_( str.data() ), n_( str.size() )
     ^~~~~~~~~~~~~~~~~
./boost/core/detail/string_view.hpp:335:26: note: candidate: 'constexpr boost::core::basic_string_view<char>::basic_string_view(const boost::core::basic_string_view<char>&)'
 template<class Ch> class basic_string_view
                          ^~~~~~~~~~~~~~~~~
./boost/core/detail/string_view.hpp:335:26: note: candidate: 'constexpr boost::core::basic_string_view<char>::basic_string_view(boost::core::basic_string_view<char>&&)'
In file included from ./boost/json/value.hpp:20,
                 from ./boost/json/array.hpp:1756,
                 from ./boost/json.hpp:15,
                 from ./boost/json/src.hpp:27,
                 from libs/json/src/src.cpp:10:
```

Full build log attached:
[boost-build.log](https://github.com/boostorg/json/files/8775149/boost-build.log)

This happens with gcc 8.3 that is shipped with Debian 10 Buster, when building in C++17 mode with GNU extensions.

vinniefalco:
fix #708


Lastique:
Comparison operators need to change as well.

Lastique:
Also, although it might fix the Boost.JSON building issue, users' code that invokes conversion to `boost::core::string_view` will still remain affected. I think, the conversion itself needs to be fixed (i.e. one of the two conversion paths should be made less preferable).

vinniefalco:
Maybe we should apply a gcc 8.3 specific fix?

vinniefalco:
I updated the comparisons

include/boost/json/string.hpp
```diff
@@ -2782,7 +2792,7 @@ typename std::enable_if<
 operator==(T const& lhs, U const& rhs) noexcept
 #endif
 {
-    return string_view(lhs) == string_view(rhs);
+    return lhs.subview() == rhs.subview();
```
Lastique, line 2795: This may not work if either `T` or `U` is neither `string` or `string_view`. You're testing for `is_convertible`, which allows any type that may not have `subview()` member.

Lastique:
> Maybe we should apply a gcc 8.3 specific fix?

I'm not opposed to a gcc-specific workaround. I've just not figured out a way to make the conversions to compile.

vinniefalco:
How about this? https://github.com/boostorg/json/pull/709/commits/1949b0abb0249eec5bd00f3b86e86dd14fedb382

vinniefalco:
It seems to compile on gcc 8.3 but I am not sure whether I need to actually instantiate those operators or not
https://godbolt.org/z/EdxzEGT4v

Lastique:
Seems to be working for my local build of Boost.JSON. Also on [godbolt](https://godbolt.org/z/9crKrMTnG). Thanks!

PR #713, opened 2022-06-03
set_at_pointer

### LINKED ISSUE #700: Inserting values at pointers
Currently trying out the new JSON pointer implementation and I think this would be very useful in many cases.

One thing I have needed is inserting or assigning arbitrary JSON values at locations specified by JSON pointers. Is it currently possible? How can I achieve it in the current state?

grisumbras:
Fix #700

This implementation makes these choices:
1. Nulls are replaced when creating intermediaries.
2. When creating intermediaries segments are always treated as object keys, (that is only objects are created as intermediaries).
3. Arrays are extended by 1 element at most.

### DIFF
--- a/include/boost/json/impl/pointer.ipp
+++ b/include/boost/json/impl/pointer.ipp
@@ -22,11 +22,12 @@ public:
     class iterator;
 
     pointer_token(
-        char const* b,
-        char const* e) noexcept
-        : b_(b)
-        , e_(e)
+        string_view sv) noexcept
+        : b_( sv.begin() + 1 )
+        , e_( sv.end() )
     {
+        BOOST_ASSERT( !sv.empty() );
+        BOOST_ASSERT( *sv.data() == '/' );
     }
 
     iterator begin() const noexcept;
@@ -142,7 +143,9 @@ bool is_invalid_zero(
     ++b;
     if( b == e )
         return false;
-    return *b != '/';
+
+    BOOST_ASSERT( *b != '/' );
+    return true;
 }
 
 bool is_past_the_end_token(
@@ -153,17 +156,22 @@ bool is_past_the_end_token(
         return false;
 
     ++b;
-    if( b == e )
-        return true;
-    return *b == '/';
+    BOOST_ASSERT( (b == e) || (*b != '/') );
+    return b == e;
 }
 
 std::size_t
 parse_number_token(
-    char const*& b,
-    char const* e,
+    string_view sv,
     error_code& ec) noexcept
 {
+    BOOST_ASSERT( !sv.empty() );
+
+    char const* b = sv.begin();
+    BOOST_ASSERT( *b == '/' );
+
+    ++b;
+    char const* const e = sv.end();
     if( ( b == e )
         || is_invalid_zero(b, e) )
     {
@@ -173,6 +181,7 @@ parse_number_token(
 
     if( is_past_the_end_token(b, e) )
     {
+        ++b;
         BOOST_JSON_FAIL(ec, error::past_the_end);
         return {};
     }
@@ -181,9 +190,7 @@ parse_number_token(
     for( ; b != e; ++b )
     {
         char const c = *b;
-
-        if( '/' == c)
-            break;
+        BOOST_ASSERT( c != '/' );
 
         unsigned d = c - '0';
         if( d > 9 )
@@ -205,13 +212,23 @@ parse_number_token(
     return result;
 }
 
-pointer_token
-get_token(
-    char const* b,
-    char const* e,
+string_view
+next_segment(
+    string_view& sv,
     error_code& ec) noexcept
 {
-    char const* start = b;
+    if( sv.empty() )
+        return sv;
+
+    char const* const start = sv.begin();
+    char const* b = start;
+    if( *b++ != '/' )
+    {
+        BOOST_JSON_FAIL( ec, error::missing_slash );
+        return {};
+    }
+
+    char const* e = sv.end();
     for( ; b < e; ++b )
     {
         char const c = *b;
@@ -222,7 +239,7 @@ get_token(
         {
             if( ++b == e )
             {
-                BOOST_JSON_FAIL(ec, error::invalid_escape);
+                BOOST_JSON_FAIL( ec, error::invalid_escape );
                 break;
             }
 
@@ -233,7 +250,7 @@ get_token(
                 // valid escape sequence
                 continue;
             default: {
-                BOOST_JSON_FAIL(ec, error::invalid_escape);
+                BOOST_JSON_FAIL( ec, error::invalid_escape );
                 break;
             }
             }
@@ -241,7 +258,8 @@ get_token(
         }
     }
 
-    return pointer_token(start, b);
+    sv.remove_prefix( b - start );
+    return string_view( start, b );
 }
 
 value*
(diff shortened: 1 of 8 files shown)


vinniefalco:
Why did performance change

grisumbras:
Mystery to me too

vinniefalco:
Kind of a long section name though
image

include/boost/json/impl/pointer.ipp
```diff
@@ -336,6 +392,69 @@ value::find_pointer(string_view ptr, std::error_code& ec) noexcept
     return const_cast<value*>(self.find_pointer(ptr, ec));
 }
 
+value&
+value::set_at_pointer(string_view ptr, value_ref jv)
```
vinniefalco, line 396: `jv` is the canonical parameter name for `value`.

`value_ref` is not that.

include/boost/json/impl/pointer.ipp
```diff
@@ -336,6 +392,69 @@ value::find_pointer(string_view ptr, std::error_code& ec) noexcept
...
+{
+    error_code ec;
+    auto const result = try_set_at_pointer(ptr, jv, ec);
+    if( ec )
+        detail::throw_system_error(ec, BOOST_JSON_SOURCE_POS);
```
vinniefalco, line 401: these parameters could be defaulted in the declaration, e.g.
```
BOOST_URL_DECL void BOOST_NORETURN throw_length_error(char const* what,
    source_location const& loc = BOOST_CURRENT_LOCATION);
```

include/boost/json/impl/pointer.ipp
```diff
@@ -336,6 +392,69 @@ value::find_pointer(string_view ptr, std::error_code& ec) noexcept
...
+            value val(object(obj.storage()), obj.storage());
+            key_value_pair kv(key, pilfer(val));
+            obj.reserve(obj.size() + 1);
+            return &obj.insert_impl(pilfer(kv)).first->value();
+        },
+        [](array& arr, std::size_t index, error_code& ec) -> value*
+        {
+            if( ec == error::past_the_end || index == arr.size() )
+            {
+                ec.clear();
```
vinniefalco, line 431: `ec = {}` is canonical

include/boost/json/value.hpp
```diff
@@ -3293,6 +3293,111 @@ class value
...
+        <br/>
+
+        The type of created intermediate elements is always @ref object. The
+        function creates at most one element for each pointer segment.
+
+        @par Complexity
+        Linear in the size of `ptr`.
+
+        @par Exception Safety
+        Basic guarantee.
```
vinniefalco, line 3328: but we could offer the Strong Guarantee, no?  If we do two passes through the pointer?

include/boost/json/value.hpp
```diff
@@ -3293,6 +3293,111 @@ class value
...
+
+        @param ptr JSON Pointer string.
+
+        @param jv The value to assign to pointed element.
+
+        @return Pointer to the element identified by `ptr`.
+
+        @see
+        <a href="https://datatracker.ietf.org/doc/html/rfc6901">
+            RFC 6901 - JavaScript Object Notation (JSON) Pointer</a>
```
vinniefalco, line 3338: I format it this way
```
        <a href="https://datatracker.ietf.org/doc/html/rfc6901"
            >JavaScript Object Notation (JSON) Pointer (rfc6901)</a>
```

vinniefalco:
If it doesn't already say so, the javadoc / exposition for `value_ref` should explain that objects of that type should only be declared as function parameters. Never as class members, local variables, or objects with static storage duration.

vinniefalco:
The decision to only create objects along the way seems unintuitive and controversial. What do other implementations of JSON Pointer do? What is the rationale for this?

include/boost/json/impl/pointer.ipp
```diff
@@ -336,6 +392,69 @@ value::find_pointer(string_view ptr, std::error_code& ec) noexcept
...
+            {
+                key_value_pair* kv = detail::find_in_object(obj, token).first;
+                if( kv )
+                    return &kv->value();
+            }
+
+            string key(token.begin(), token.end(), obj.storage());
+            value val(object(obj.storage()), obj.storage());
+            key_value_pair kv(key, pilfer(val));
+            obj.reserve(obj.size() + 1);
```
vinniefalco, line 424: `insert_impl` already reserves first, you dont need to do it

vinniefalco:
This PR is not ready... needs work :)

include/boost/json/impl/pointer.ipp
vinniefalco: What do you think of this
```
    object tmp( { { string_view(token), object_kind } }, obj.storage() );
    obj.swap( tmp );
    return obj.begin()
```

doc/qbk/03_08_nested_access.qbk
```diff
@@ -9,7 +9,7 @@
 
 [/-----------------------------------------------------------------------------]
 
-[section:nested_access Accessing Deeply Nested Elements]
+[section:nested_access Accessing and Changing Deeply Nested Elements]
```
vinniefalco, line 12: Quite a long section name this will appear in the toc?

doc/qbk/03_08_nested_access.qbk
```diff
@@ -26,4 +26,14 @@ with
...
 [snippet_pointer_3]
 
+The library also allows changing and adding deeply nested elements:
+
+[snippet_pointer_4]
+
+The function __set_at_pointer__ traverses the value similarly to
+__at_pointer__, but additionally, if it encounters a `null` value it replaces
+it with an __object__, and when the referenced object or __array__ element
+doesn't exist, it creates an object. A version of that function that reports
```
vinniefalco, line 36: That's a lot of commas

include/boost/json/pointer.hpp
```diff
@@ -0,0 +1,30 @@
...
+#ifndef BOOST_JSON_POINTER_HPP
+#define BOOST_JSON_POINTER_HPP
+
+#include <boost/json/string.hpp>
+#include <boost/json/system_error.hpp>
+
+BOOST_JSON_NS_BEGIN
+
+std::pair<std::size_t, string_view>
+parse_pointer_number_segment(
```
vinniefalco, line 19: This is the name of a public function?

include/boost/json/value.hpp
```diff
@@ -33,6 +33,11 @@
 
 BOOST_JSON_NS_BEGIN
 
+template< bool Const >
+struct basic_pointer_traverse_state;
+using pointer_traverse_state = basic_pointer_traverse_state< false >;
+using const_pointer_traverse_state = basic_pointer_traverse_state< true >;
```
vinniefalco, line 39: These are public?

include/boost/json/value.hpp
```diff
@@ -3266,6 +3271,121 @@ class value
...
+    try_set_at_pointer(string_view ptr, value_ref jv, std::error_code& ec);
+/** @} */
+
+    //------------------------------------------------------
+
+    BOOST_JSON_DECL
+    const_pointer_traverse_state
+    step_over_pointer(string_view ptr) const noexcept;
+
+    inline
```
vinniefalco, line 3383: the `inline` goes on the definition not the declaration

include/boost/json/value.hpp
```diff
@@ -3266,6 +3271,121 @@ class value
...
+    BOOST_JSON_DECL
+    value*
+    try_set_at_pointer(string_view ptr, value_ref jv, std::error_code& ec);
+/** @} */
+
+    //------------------------------------------------------
+
+    BOOST_JSON_DECL
+    const_pointer_traverse_state
+    step_over_pointer(string_view ptr) const noexcept;
```
vinniefalco, line 3381: `ptr` is misleading, the convention in my libraries is to name the variable based on its type:

`jv` json::value
`s` string or string_view
`ec` error code

Only very rarely are things named after what they are, e.g. `string_view key`. But `ptr` is very suggestive of a pointer which it is not. This is misleading

include/boost/json/pointer.hpp
grisumbras: Yes. It's a helper that users need to implement their own behaviour while traversing pointers.

include/boost/json/value.hpp
grisumbras: Yes

test/pointer.cpp
```diff
@@ -8,12 +8,104 @@
...
+        if( state.result )
+        {
+            tgt = state.result;
+            sv.remove_prefix( state.token.size() );
+            BOOST_ASSERT( !state.error.failed() );
+            continue;
+        }
+
+        assert( state.error.failed() );
+        if( state.error == json::error::past_the_end )
```
vinniefalco, line 40: Why isn't this simply `error::end`?

pdimov (approved):
This is what we've been talking about, yes. Everything looks fine at a quick glance.

PR #719, opened 2022-06-09
Fix/deduce endianness

### LINKED ISSUE #717: bug on solaris
When parsing the int field information in JSON message on Solaris, the value obtained is incorrect. The order of the numbers is chaotic.
Here are some codes.
 "recvMsg " is a JSON string that needs to be parsed.
```c++
boost::json::object recvJson = boost::json::parse(recvMsg.c_str()).as_object();
boost::json::array playSequence = recvJson["playSequence"].as_array();
for (boost::json::value info : playSequence)
{
    boost::json::object infoJson = info.as_object();
    long long cameraId = infoJson["cameraNumber"].as_int64();
    printf("cameraid : %lld\n",cameraId);
}

```
The correct value is 11739859, but the actual value is 37119589. The order of numbers has changed.
We tested normally on windows10 (with MSVC v143), but there was a problem on Solaris 11.4.42.111.0.(sparcv9 ,with GCC v11.2.0).
We tested 1_ 78_ 0 and 1_ 79_ 0, their performance is the same.

grisumbras:
Fix #717 
Also see #576

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -14,8 +14,8 @@ windowsglobalimage="cppalliance/dronevs2019"
 
 def main(ctx):
   return [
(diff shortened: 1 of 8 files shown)


vinniefalco:
Please no negative commit messages e.g. "confusing code." Instead express it as a positive "add clarifying implementation notes"

PR #727, opened 2022-06-30
fix handling of comment after trailing comma

### LINKED ISSUE #726: trailing commas and comments not working well in some cases
### Version of Boost
1.79.0

``` c++
#include <boost/json/parse.hpp>
#include <boost/json/value.hpp>

int main(int argc, char* argv[]) {
    std::string raw = R"([
    "123", // 456
])";
    boost::json::parse_options opt {};
    opt.allow_comments = true;
    opt.allow_trailing_commas = true;
    boost::json::value val = boost::json::parse(raw, {}, opt);
    return 0;
}

```

compile :
``` bash
g++ -I /data/vendor/boost-1.79/include -L /data/vendor/boost-1.79/lib test.cpp -lboost_json -lboost_system -o test_json
```

run the program, and it will terminate with the following: 
``` bash
terminate called after throwing an instance of 'boost::wrapexcept<boost::system::system_error>'
  what():  syntax error [boost.json:1 at ./boost/json/basic_parser_impl.hpp:634 in function 'parse_value']
```

either remove the comment ("// 456") or the trailing commas, the exception will go away.

gcc/g++ 7.3.1

grisumbras:
Fix #726

### DIFF
--- a/include/boost/json/basic_parser_impl.hpp
+++ b/include/boost/json/basic_parser_impl.hpp
@@ -1894,6 +1894,7 @@ do_arr1:
         return maybe_suspend(cs.begin(), state::arr1, size);
     if(BOOST_JSON_LIKELY(*cs != ']'))
     {
+loop:
         if(allow_comments && *cs == '/')
         {
 do_arr2:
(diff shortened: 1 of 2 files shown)


include/boost/json/basic_parser_impl.hpp
```diff
@@ -1894,6 +1894,7 @@ parse_array(const char* p,
         return maybe_suspend(cs.begin(), state::arr1, size);
     if(BOOST_JSON_LIKELY(*cs != ']'))
     {
+loop:
```
vinniefalco, line 1897: nice

PR #741, opened 2022-08-24
Feature/noexcept conversions

### LINKED ISSUE #667: Motivating use case for nonthrowing value_to
Suppose that a user wants to represent `variant<T1, T2, T3>` as just the active alternative; that is, when the variant contains `T1`, its `value_from` function just forwards to T1's `value_from`, and similarly for T2 and T3. When converting from a `value`, it tries the alternatives in order, and uses the first one that matches. (https://godbolt.org/z/GqKPYx4Tq)

This occurs in practice when the JSON schema is externally specified and the types of some values can vary. For instance, Boost's `meta/libraries.json` allows both a string and an array of strings in the `authors` and `maintainers` fields. This can be represented by `variant<string, vector<string>>`.

The solution in the above CE link works, but it throws `sizeof...(T) / 2` times on average on each `value_to` call, which kills performance. It would be much better if a nonthrowing alternative of `value_to` existed that returned the parse/type errors using `error_code`. In fact, the code above emulates this nonthrowing alternative with the `try_value_to` function.

There are still few users for the `tag_invoke` mechanism so switching to a nonthrowing support function _now_ won't break much existing code, although the longer we wait, the less true this becomes.

### LINKED ISSUE #736: Add a nonthrowing overload for value_to
Implementing `value_to` for `variant` (for one possible representation) involves repeatedly trying `value_to` for each alternative. Since `value_to` reports failures via exceptions, this will be _slow_. It would be better if `value_to` had a `noexcept` overload that could be used for this purpose.

There's no perfect option here, because returning `result<T>` would impose a dependency on System, and we want people to be able to add JSON support without a dependency. The other alternative
```
bool value_to( value const& jv, T& t, error_code& ec );
```
would require `T` to be default-constructible, but this may be good enough, as the "traditional" `value_to` can be used for other types.

A default implementation for this overload would be

```
bool value_to( value const& jv, T& t, error_code& ec )
{
    try
    {
        t = value_to( jv );
        return true;
    }
    catch( system_error const& x )
    {
        ec = x.code();
    }
    catch( ... )
    {
        ec = ...;
    }

    return false;
}
```

but "basic" types should have proper nonthrowing overloads.

grisumbras:
Fix #736
Fix #667

### DIFF
--- a/include/boost/json/conversion.hpp
+++ b/include/boost/json/conversion.hpp
@@ -11,6 +11,7 @@
 #define BOOST_JSON_CONVERSION_HPP
 
 #include <boost/json/string_view.hpp>
+#include <boost/json/value.hpp>
 #include <boost/mp11/algorithm.hpp>
 #include <boost/mp11/utility.hpp>
 
@@ -67,6 +68,53 @@ template<class T>
 using is_value_type_pair = is_value_type_pair_helper<value_type<T>>;
 
 } // namespace detail
+
+/**
+   Helper trait that returns @ref result
+
+   The primary template is an incomplete type. The library provides a partial
+   specialisation `result_for<T1, value>`, that has nested type alias `type`
+   that aliases the type `result<T1>`.
+
+   The purpose of this trait is to let users provide non-throwing conversions
+   for their types without creating a physical dependency on Boost.Json. For
+   example:
+
+   @code
+   namespace boost
+   {
+   namespace json
+   {
+
+   template<class T>
+   struct value_to_tag;
+
+   template<class T1, class T2>
+   struct result_for;
+   }
+   }
+
+   namespace mine
+   {
+       class my_class;
+       ...
+       template<class JsonValue>
+       boost::json::result_for<my_class, JsonValue>
+       tag_invoke(boost::json::try_value_to_tag<my_class>, const JsonValue& jv)
+       { ... }
+   }
+   @endcode
+
+    @see @ref try_value_to, @ref try_value_to_tag
+*/
+template <class T1, class T2>
+struct result_for;
+
+template <class T>
+struct result_for<T, value>
+{
+    using type = result< detail::remove_cvref<T> >;
+};
 
 /** Determine if `T` can be treated like a string during conversions.
 
--- a/include/boost/json/error.hpp
+++ b/include/boost/json/error.hpp
@@ -101,6 +101,28 @@ enum class error
 
     /// past-the-end index is not supported
     past_the_end,
+
+    //
+    // Conversion errors
+    //
+
+    /// JSON null was expected during conversion
+    not_null,
+
+    /// JSON bool was expected during conversion
+    not_bool,
+
+    /// JSON array was expected during conversion
+    not_array,
+
+    /// JSON object was expected during conversion
+    not_object,
+
+    /// JSON string was expected during conversion
+    not_string,
+
+    /// JSON array has size incompatible with target
+    size_mismatch,
 };
 
 /** Error conditions corresponding to JSON errors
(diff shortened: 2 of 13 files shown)


include/boost/json/error.hpp
```diff
@@ -101,6 +101,16 @@ enum class error
...
+
+    //
+    // Conversion errors
+    //
+
+    /// JSON bool was expected during conversion
+    bool_expected,
+
+    /// JSON string was expected during conversion
+    string_expected,
```
vinniefalco, line 113: when you add errors these need to be added to error.ipp and error.cpp (tests) as well

include/boost/json/system_error.hpp
```diff
@@ -28,6 +29,9 @@ using error_condition = boost::system::error_condition;
 /// The type of system error thrown by the library.
 using system_error = boost::system::system_error;
 
+template< class T >
+using result = boost::system::result<T>;
```
vinniefalco, line 33: Boost.JSON should follow the model used in Boost.URL. You can probably just copy this file:
https://github.com/CPPAlliance/url/blob/develop/include/boost/url/error_types.hpp

and change the existing JSON headers to include this file instead. Keep the original JSON headers if they overlap and just edit them to only include this new file.

include/boost/json/error.hpp
grisumbras: Don't review this, it's just an outline

include/boost/json/system_error.hpp
```diff
@@ -28,6 +29,9 @@ using error_condition = boost::system::error_condition;
 /// The type of system error thrown by the library.
 using system_error = boost::system::system_error;
 
+template< class T >
```
vinniefalco, line 32: You have to put a javadoc on this and add it to the reference :) See next note

test/doc_forward_conversion.cpp
```diff
@@ -93,6 +176,15 @@ class doc_forward_conversion
 
         value const jv2 = value_from( c );
         BOOST_TEST( jv == jv2 );
+
+//[doc_forward_conversion_3_2
+        auto const rc = try_value_to<thirdparty::customer>( jv );
```
vinniefalco, line 181: we should avoid `auto` in examples

test/doc_forward_conversion.cpp
```diff
@@ -93,6 +176,15 @@ class doc_forward_conversion
...
+        BOOST_TEST( rc.has_value() );
+
+        auto const rv = try_value_from( *rc );
+        BOOST_TEST( rv.has_value() );
```
vinniefalco, line 185: Are you showing `BOOST_TEST` in examples?

include/boost/json/conversion.hpp
```diff
@@ -68,6 +69,53 @@ using is_value_type_pair = is_value_type_pair_helper<value_type<T>>;
...
+   that aliases the type `result<T1>`.
+
+   The purpose of this trait is to let users provide non-throwing conversions
+   for their types without creating a physical dependency on Boost.Json. For
+   example:
+
+   @code
+   namespace boost
+   {
+   namespace jsonj
```
pdimov, line 86: Typo? jsonj

grisumbras: Yes, it was a typo

include/boost/json/system_error.hpp
```diff
@@ -28,6 +29,69 @@ using error_condition = boost::system::error_condition;
...
+        // Return true if the result contains an error
+        constexpr bool has_error() const noexcept;
+
+        // These two return true if the result contains a value
+        constexpr bool has_value() const noexcept;
+        constexpr explicit operator bool() const noexcept;
+
+        // Return the value or throw an exception if has_value() == false
+        constexpr T& value();
+        constexpr T& operator*();
```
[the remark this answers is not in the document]
grisumbras: Reorganised

Issue #748, opened 2022-08-28
Support for ordered insert/erase on json::object


gummif:
I have a json that needs to be manipulated and maintaining the order of the keys in the object is really imported (even though it is supposed to be unordered key-value map).

Currently insertions are ordered so creating a json::object and inserting at the end will preserve the order. A work around I have now is something like this:
```c++
template<class K, class P>
void ordered_insert_before(boost::json::object& o, K&& key, P&& p)
{
	auto x = boost::json::object(o.storage());
	x.reserve(o.size() + 1);
	bool added = false;
	for (auto&& [k, v] : o)
	{
		if (k == key && !added)
		{
			x.insert(std::forward<P>(p));
			added = true;
		}
		x.insert(boost::json::object::value_type{std::move(k), std::move(v)});
	}
	if (!added)
	{
		x.insert(std::forward<P>(p));
	}
	o.swap(x);
}

template<class K>
void ordered_erase(boost::json::object& o, K&& key)
{
	auto x = boost::json::object(o.storage());
	x.reserve(o.size());
	for (auto&& [k, v] : o)
	{
		if (k != key)
		{
			x.insert(boost::json::object::value_type{std::move(k), std::move(v)});
		}
	}
	o.swap(x);
}
```

There are of course O(N). Is there interest for adding these as member functions or free functions with stronger ordering guarantees but worse O(N) complexity + allocations?

vinniefalco:
In order to `memmove` the elements in the `object` correctly, the implementation would have to calculate the hash of each moved element, and adjust its index in the hash table.

PR #749, opened 2022-08-29
object::stable_erase

gummif:
Support for erasing an element without reordering. This is part one of support https://github.com/boostorg/json/issues/748, part two would be stable_insert.

### DIFF
--- a/include/boost/json/impl/object.ipp
+++ b/include/boost/json/impl/object.ipp
@@ -508,30 +508,17 @@ object::
 erase(const_iterator pos) noexcept ->
     iterator
 {
-    auto p = begin() + (pos - begin());
-    if(t_->is_small())
-    {
-        p->~value_type();
-        --t_->size;
-        auto const pb = end();
-        if(p != end())
-        {
+    return do_erase(pos,
+        [this](iterator p) {
             // the casts silence warnings
             std::memcpy(
                 static_cast<void*>(p),
-                static_cast<void const*>(pb),
+                static_cast<void const*>(end()),
                 sizeof(*p));
-        }
-        return p;
-    }
-    remove(t_->bucket(p->key()), *p);
-    p->~value_type();
-    --t_->size;
-    if(p != end())
-    {
-        reindex_relocate(end(), p);
-    }
-    return p;
+        },
+        [this](iterator p) {
+            reindex_relocate(end(), p);
+        });
 }
 
 auto
@@ -551,30 +538,20 @@ object::
 stable_erase(const_iterator pos) noexcept ->
     iterator
 {
-    auto p = begin() + (pos - begin());
-    if(t_->is_small())
-    {
-        p->~value_type();
-        --t_->size;
-        if(p != end())
-        {
+    return do_erase(pos,
+        [this](iterator p) {
             // the casts silence warnings
             std::memmove(
                 static_cast<void*>(p),
                 static_cast<void const*>(p + 1),
                 sizeof(*p) * (end() - p));
-        }
-        return p;
-    }
-    remove(t_->bucket(p->key()), *p);
-    p->~value_type();
-    --t_->size;
-    auto pret = p;
-    for (; p != end(); ++p)
-    {
-        reindex_relocate(p + 1, p);
-    }
-    return pret;
+        },
+        [this](iterator p) {
+            for (; p != end(); ++p)
+            {
+                reindex_relocate(p + 1, p);
+            }
+        });
 }
 
 auto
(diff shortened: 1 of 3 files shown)


vinniefalco:
impressive work :) you figured out how `object` is implemented...

include/boost/json/object.hpp
```diff
@@ -1166,6 +1166,51 @@ class object
     std::size_t
     erase(string_view key) noexcept;
 
+    /** Erase an element preserving order
+
+        Remove the element pointed to by `pos`, which must
+        be valid and dereferenceable. Thus the @ref end()
+        iterator (which is valid but cannot be dereferenced)
+        cannot be used as a value for `pos`.
+        All references and iterators are invalidated.
```
vinniefalco, line 1175: Is this true? Or does it only invalidate iterators from pos and later?

gummif: However for the key erase function I don't see a reason to be more specific since we are not passing in an iterator.

vinniefalco: agreed

vinniefalco:
@grisumbras I guess I have no objection to offering this feature. If the user wants to opt-in to preserving order at a computational cost, I can't see a downside. But its up to you, maybe there's something I haven't thought of? What does Peter think?

include/boost/json/object.hpp
```diff
@@ -1567,6 +1614,12 @@ class object
     destroy(
         key_value_pair* first,
         key_value_pair* last) noexcept;
+
+    inline
```
sehe, line 1618: Could this use a little javadoc even though it's internal?

include/boost/json/object.hpp
```diff
@@ -1166,6 +1166,53 @@ class object
...
+        are not invalidated.
+        The relative order of non-erased elements is preserved.
+
+        @par Complexity
+        Linear in @ref size().
+
+        @par Exception Safety
+        No-throw guarantee.
+
+        @return An iterator following the last removed element.
```
sehe, line 1186: "last" seems redundant here

include/boost/json/object.hpp
```diff
@@ -1166,6 +1166,53 @@ class object
     std::size_t
     erase(string_view key) noexcept;
 
+    /** Erase an element preserving order
+
+        Remove the element pointed to by `pos`, which must
+        be valid and dereferenceable. Thus the @ref end()
+        iterator (which is valid but cannot be dereferenced)
+        cannot be used as a value for `pos`.
```
sehe, line 1174: The sentences "Thus the @ref end() iterator (which is valid but cannot be dereferenced) cannot be used as a value for `pos`." is redundant.

If this is expected to be a gotcha (?) I'd move it to a note below.

sehe (commented):
Got nerd sniped by this PR. Nice work :)

include/boost/json/object.hpp
vinniefalco: eh.. 


by the way docca supports `@param firs,t last The range to destroy` now

include/boost/json/object.hpp
gummif: It looks like this is verbatim from https://en.cppreference.com/w/cpp/container/vector/erase. Moved to note section.

include/boost/json/impl/object.ipp
```diff
@@ -556,6 +546,49 @@ erase(string_view key) noexcept ->
     return 1;
 }
 
+auto
+object::
+stable_erase(const_iterator pos) noexcept ->
```
grisumbras, line 551: `stable_erase` and `erase` share quite a lot of code. A feel like a better approach is having a private  function template `iterator do_erase(const_iterator pos, F get_pb)` that ends with
```c++
auto result = p;
while( p != end() )
{
    auto pb = get_pb(p);
    reindex_relocate(pb, p);
    p = pb;
}
return result;
```
`erase` would be implemented like `return do_erase(pos, [this](iterator) { return end(); })`
`stable_erase` would be implemented like `return do_erase(pos, [](iterator p) { return ++p; })`

include/boost/json/object.hpp
```diff
@@ -1166,6 +1169,56 @@ class object
...
     erase(string_view key) noexcept;
 
+    /** Erase an element preserving order
+
+        Remove the element pointed to by `pos`, which must
+        be valid and dereferenceable.
+        References and iterators from `pos` to `end()`, both
+        included, are invalidated. Other iterators and references
+        are not invalidated.
+        The relative order of non-erased elements is preserved.
```
grisumbras, line 1179: s/non-erased/remaining/

include/boost/json/object.hpp
```diff
@@ -1166,6 +1169,56 @@ class object
...
+    */
+    BOOST_JSON_DECL
+    iterator
+    stable_erase(const_iterator pos) noexcept;
+
+    /** Erase an element preserving order
+
+        Remove the element which matches `key`, if it exists.
+        All references and iterators are invalidated.
+        The relative order of non-erased elements is preserved.
```
grisumbras, line 1205: s/non-erased/remaining/

include/boost/json/impl/object.ipp
sehe: I looked at the same a while back. A subtle difference (easy to miss) is that memcpy is memmov here and acts on multiple elements, instead of just 1. I still feel that it can probably be a shared implementation, but it require some scrutiny.

gummif: This has been refactored to a common implementation.

vinniefalco: For the record I was fine with it before the refactoring

grisumbras:
Thanks for the feature.

Issue #754, opened 2022-09-20
Insert object into an existing object as a key-value-pair


Tabascl:
Boost Version 1.80
I have a class hierarchy that looks like this: class A is composed of classes B and C.
B and C each have a function that returns boost::json::object, which looks like this in case of B: `{"B": "someValue"}`

A is supposed to compose each of these objects into a structure like this:
```
{
  "A": {
    "B": "someValue",
    "C": "someOtherValue"
  }
}
```

The problem I'm facing is, that the resulting object always looks like this:
```
{
  "A": [
    { "B": "someValue" },
    { "C": "someOtherValue" }
  ]
}
```
As you can see, the objects from the subclasses are not added as key-value pairs, but as objects, each containing a single pair, into an array.

Here's a snippet to reproduce the issue:
```
#include <iostream>
#include <boost/json.hpp>

using namespace boost::json;

object makeObject(std::string key, int value)
{
    object obj;
    obj[key] = value;
    return obj;
}

int main() {
    object obj;
    obj["main"] = {makeObject("test1", 123), makeObject("test2", 456)};

    std::cout << obj << std::endl;
    return 0;
}
```

According to the Quick Look documentation, I would have expected this to not create an array with single-value objects in it.
What am I doing wrong here?

grisumbras:
The correct incantation in this case should probably be
```c++
{ {"test1", 123}
, {"test2", 456}
}
```

Tabascl:
Maybe I should have been more clear here, I sadly can't do the incantation like you said. The makeObject function is here to enable a very simple example that shows the problem.
Here's a more complete example, that hopefully outlines my problem a bit better:
```
#include <boost/json.hpp>
#include <iostream>

using namespace boost::json;

class B
{
public:
    object makeProp()
    {
        object obj;
        obj["B"] = {{"property1", 1}};
        return obj;
    }
};

class C
{
public:
    object makeProp()
    {
        object obj;
        obj["C"] = {{"property1", 1}, {"property2", {1, 2, 3}}};
        return obj;
    }
};

class A
{
public:
    object makeProp()
    {
        B b;
        C c;

        object obj;
        obj["A"] = {b.makeProp(), c.makeProp()};
        return obj;
    }
};

int main()
{
    A a;
    std::cout << a.makeProp() << std::endl;

    return 0;
}
```
Hopefully the purpose of the makeObject-function in the original example is clearer now. I somehow need to insert objects created elsewhere directly into another one.

[...]

PR #755, opened 2022-09-21
Add merge examples

vinniefalco:
proposed for #754

### DIFF
--- a/example/merge.cpp
+++ b/example/merge.cpp
@@ -1,0 +1,129 @@
+//
+// Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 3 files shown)


example/merge.cpp
```diff
@@ -0,0 +1,129 @@
...
+    while(it != begin)
+    {
+        --it;
+        auto result =
+            dest.insert(pilfer(*it));
+        if(result.second)
+            src.erase(it);
+    }
+    return dest;
+}
```
grisumbras, line 57: This should be almost equivalent to your code:
```c++
object&
merge_objects(
    object& dest,
    object src)
{
    dest.insert(
        std::make_move_iterator(src.begin()),
        std::make_move_iterator(src.end()));
    return dest;
}
```
And it's basically one library function (with a little help of the standard library). `merge_copy` is the same function, only the caller capies rather than moves `src`.

vinniefalco:
NB: I'm not suggesting this be merged, its up to you to decide what to do with it (which could be, to do nothing)

PR #769, opened 2022-10-14
value_to/from supports std::optional

### LINKED ISSUE #677: Does it support std::optional or boost::optional out of the box ?
Hello,

Does it support simple json conversion from `std::optional<std::string>` for instance ? When I test I get some errors.
Thanks

grisumbras:
Fix #677


include/boost/json/detail/value_from.hpp
```diff
@@ -148,7 +152,50 @@ value_from_helper(
...
+        jv = nullptr;
+}
+
+inline
+void
+tag_invoke(
+    value_from_tag,
+    value& jv,
+    std::nullopt_t)
+{
```
pdimov, line 192: Shouldn't `nullopt_t` be marked as `is_null_like` instead?

include/boost/json/detail/value_from.hpp
```diff
@@ -148,7 +152,50 @@ value_from_helper(
...
+    detail::is_optional< detail::remove_cvref<T> >::value
+    >::type
+tag_invoke(
+    value_from_tag,
+    value& jv,
+    T&& from)
+{
+    using InnerRef = decltype( *static_cast<T>(from) );
+    if( from )
+        value_from(static_cast<InnerRef>(*from), jv);
```
pdimov, line 181: Why is this `static_cast` necessary? Why not just `*from`?

include/boost/json/detail/value_from.hpp
grisumbras: The generic implementation does not work, because std::nullopt_t is not default constructible.

PR #791, opened 2022-10-27
serializer can work on user-defined types directly

vinniefalco:
This is just an experimental draft


include/boost/json/detail/impl/writer.ipp
```diff
@@ -0,0 +1,27 @@
...
+
+#ifndef BOOST_JSON_DETAIL_IMPL_WRITER_IPP
+#define BOOST_JSON_DETAIL_IMPL_WRITER_IPP
+
+#include <boost/json/detail/writer.hpp>
+#include <boost/json/detail/sse2.hpp>
+#include <boost/json/detail/format.hpp> // for max_number_chars
+#include <boost/static_assert.hpp>
+
+namespace boost {
```
vinniefalco, line 19: we should probably get rid of this file :)

PR #792, opened 2022-10-27
test conversion to string_view

vinniefalco:


grisumbras:
This is a problem with GCC 8.3 that doesn't have a solution on our side (users will have to upgrade). In our code we should just avoid using conversion from `string` to `string_view` and use `string::subview` instead.

PR #800, opened 2022-11-05
refactor parse(istream&, ec&)

grisumbras:


vinniefalco:
should this return `result<value>`?

PR #801, opened 2022-11-06
option to check duplicate object keys

### LINKED ISSUE #127: Request for I-JSON (RFC 7493) Parsing Mode
I was wondering if there is any interest in implementing the I-JSON parsing mode for boost.json. [I-JSON](https://tools.ietf.org/html/rfc7493) is an interoperable version of JSON which aims to remove some ambiguities in the parsing. The changes required for such a compliance are listed in [Section 2](https://tools.ietf.org/html/rfc7493#section-2) of the document.

grisumbras:
Fix #127

### DIFF
--- a/include/boost/json/detail/impl/handler.ipp
+++ b/include/boost/json/detail/impl/handler.ipp
@@ -51,8 +51,13 @@ bool
 handler::
 on_object_end(
     std::size_t n,
-    error_code&)
+    error_code& ec)
 {
+    if( !ignore_duplicate_keys )
+        ec = st.check_duplicates(n);
+    if( ec.failed() )
+        return false;
+
     st.push_object(n);
     return true;
 }
--- a/include/boost/json/detail/impl/handler.ipp
+++ b/include/boost/json/detail/impl/handler.ipp
@@ -53,13 +53,8 @@ on_object_end(
     std::size_t n,
     error_code& ec)
 {
-    if( !ignore_duplicate_keys )
-        ec = st.check_duplicates(n);
-    if( ec.failed() )
-        return false;
-
-    st.push_object(n);
-    return true;
+    ec = st.push_object(n, ignore_duplicate_keys);
+    return !ec.failed();
 }
 
 bool
@@ -90,7 +85,7 @@ on_key_part(
     st.push_chars(s);
     return true;
 }
-        
+
 bool
 handler::
 on_key(
@@ -101,12 +96,12 @@ on_key(
     st.push_key(s);
     return true;
 }
-        
+
 bool
 handler::
 on_string_part(
     string_view s,
-    std::size_t, 
+    std::size_t,
     error_code&)
 {
     st.push_chars(s);
@@ -117,7 +112,7 @@ bool
 handler::
 on_string(
     string_view s,
-    std::size_t, 
+    std::size_t,
     error_code&)
 {
     st.push_string(s);
@@ -143,7 +138,7 @@ on_int64(
     st.push_int64(i);
     return true;
 }
-        
+
 bool
 handler::
 on_uint64(
@@ -165,7 +160,7 @@ on_double(
     st.push_double(d);
     return true;
 }
-        
+
 bool
 handler::
 on_bool(
@@ -192,7 +187,7 @@ on_comment_part(
 {
     return true;
 }
-        
+
 bool
 handler::
 on_comment(
--- a/include/boost/json/detail/object.hpp
+++ b/include/boost/json/detail/object.hpp
@@ -28,30 +28,29 @@ class unchecked_object
     // first one is a string key,
     // second one is the value.
     value* data_;
-    std::size_t size_;
+    value* end_;
     storage_ptr const& sp_;
+    bool ignore_duplicates_;
 
 public:
     inline
     ~unchecked_object();
 
+    inline
     unchecked_object(
         value* data,
         std::size_t size, // # of kv-pairs
-        storage_ptr const& sp) noexcept
-        : data_(data)
-        , size_(size)
-        , sp_(sp)
-    {
-    }
+        storage_ptr const& sp,
+        bool ignore_duplicates) noexcept;
 
     unchecked_object(
         unchecked_object&& other) noexcept
         : data_(other.data_)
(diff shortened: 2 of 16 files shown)


include/boost/json/detail/impl/handler.ipp
```diff
@@ -51,8 +51,13 @@ bool
 handler::
 on_object_end(
     std::size_t n,
-    error_code&)
+    error_code& ec)
 {
+    if( !ignore_duplicate_keys )
+        ec = st.check_duplicates(n);
+    if( ec.failed() )
```
vinniefalco, line 58: you check `ec` even if `ignore_duplicate_keys` is true?

grisumbras: Yeah, it's unnecessary

vinniefalco:
yeah please don't merge this with the commit message as "different approach" :)

include/boost/json/impl/object.ipp
```diff
@@ -232,6 +230,11 @@ object(detail::unchecked_object&& uo)
                 continue;
             }
             // handle duplicate
+            if( !uo.ignore_duplicate_keys() )
```
vinniefalco, line 233: This is a constant in the loop, so you should lift the conditional out of the loop

include/boost/json/impl/object.ipp
```diff
@@ -270,6 +274,11 @@ object(detail::unchecked_object&& uo)
             }
 
             // handle duplicate
+            if( !uo.ignore_duplicate_keys() )
```
vinniefalco, line 277: This is a constant in the loop, so you should lift the conditional out of the loop

include/boost/json/value_stack.hpp
```diff
@@ -355,8 +356,8 @@ class value_stack
         top of the stack to form the array.
     */
     BOOST_JSON_DECL
-    void
-    push_object(std::size_t n);
+    error_code
+    push_object(std::size_t n, bool ignore_duplicates = true);
```
vinniefalco, line 360: I think I prefer it as `allow_duplicates`, defaulting to `false`. Rationale: The default settings should have the all-zero bit-pattern. It is also easier to understand this way, you are opting in to the setting by assigning `true`. Rather than opting out to disallowing duplicates by assigning `false`.

include/boost/json/detail/value.hpp
```diff
@@ -269,6 +269,30 @@ struct access
...
+    void
+    grow_size(Object& obj) noexcept
+    {
+        ++obj.t_->size;
+    }
+
+    template< class Object >
+    static constexpr
+    index_t
+    null_index(Object const&) noexcept
```
vinniefalco, line 292: why are we mucking about with `value`?

include/boost/json/impl/object.hpp
```diff
@@ -518,19 +518,43 @@ namespace detail {
...
 ~unchecked_object()
 {
-    if(! data_)
-        return;
     if(sp_.is_not_shared_and_deallocate_is_trivial())
         return;
-    value* p = data_;
-    while(size_--)
+
+    for( value* p = data_; p != end_; p += 2 )
```
vinniefalco, line 524: Why did you change this loop? I prefer the `while` form, as it puts each of the statements on their own line. The `for` is wider, and puts multiple statements on one line, which makes debugging and code coverage worse.

include/boost/json/value_stack.hpp
grisumbras: "Ignore duplicates" and "allow duplicates" are the same thing (at least in this case). The reverse of "ingore" would be "forbid". So, "forbid_duplicate_keys" option, with `false` as default?

include/boost/json/impl/object.hpp
vinniefalco: I see why, because the loop is simpler.

include/boost/json/detail/object.hpp
```diff
@@ -60,19 +59,25 @@ class unchecked_object
...
-    release() noexcept
+    front() noexcept
     {
-        auto const data = data_;
-        data_ = nullptr;
-        return data;
+        return data_;
     }
+
+    inline
```
vinniefalco, line 78: the `inline` goes on the definition not the declaration (my fault)

include/boost/json/impl/object.ipp
```diff
@@ -70,6 +70,75 @@ find_in_object<string_view>(
     object const& obj,
     string_view key) noexcept;
 
+template< bool SmallTable, bool IgnoreDuplicates >
+void init_from_unchecked( object& obj, unchecked_object& uo )
```
vinniefalco, line 74: Should this be `noexcept`?

include/boost/json/impl/object.ipp
```diff
@@ -70,6 +70,75 @@ find_in_object<string_view>(
...
+{
+    // insert all elements, keeping
+    // the last of any duplicate keys, unless IgnoreDuplicates is false.
```
vinniefalco, line 77: There should be an assert here, the precondition of the function requires `capacity() >= uo.size()`

include/boost/json/impl/value_stack.ipp
```diff
@@ -364,16 +364,25 @@ push_array(std::size_t n)
     st_.exchange(std::move(ua));
 }
 
-void
+error_code
```
vinniefalco, line 367: this is going to force the construction of a new error code for every object even when there is no error and even when ignore_duplicates is true. It should be an outparam. Or even a bool

include/boost/json/impl/value_stack.ipp
```diff
@@ -364,16 +364,25 @@ push_array(std::size_t n)
...
-    st_.exchange(std::move(uo));
+        st_.release(n * 2), n, sp_, ignore_duplicates);
+    st_.exchange(uo);
+
+    error_code ec;
+    // constructed object should have consumed all of uo's data
+    if( uo.size() )
+    {
+        BOOST_JSON_FAIL( ec, error::duplicate_key );
+    }
```
vinniefalco, line 384: I believe that the way this is implemented has impacted performance, which is causing the regression in the per-commit performance checker.

PR #807, opened 2022-11-16
value_to supports missing elements for std::optional

grisumbras:


klemens-morgenstern:
I could also use this with `boost::optional` - here's my use-case

https://github.com/klemens-morgenstern/requests/blob/develop/example/github-issues.hpp#L94

It's quite common in APIs, that `null` and non-existant are functionally the same. 

I guess one could use a `optional<variant<monostate, T>>` if those two cases have to be distinguished.

grisumbras:
We probably need a special category for nullable types

PR #814, opened 2022-11-21
basic_parser_impl: make sentinel() return a unique pointer

MaxKellermann:
Right now, sentinel() casts the `basic_parser_impl` pointer (`this`) to `const char *`, but that pointer is not unique if the input buffer happens to be placed right before the `basic_parser_impl` instance - the end of that buffer then has the same address as `basic_parser_impl`.

Example code:

```
  struct {
    char buffer[8]{"{\"12345\""};
    boost::json::stream_parser p;
  } s;
  s.p.write(s.buffer, sizeof(s.buffer));
  s.p.write(":0}", 3);
```

This stops parsing at the end of the buffer, and then the `incomplete()` check in `parse_string()` will return true; the second `write()` call will crash with assertion failure:

>  boost/json/basic_parser_impl.hpp:1016: const char* boost::json::basic_parser<Handler>::parse_unescaped(const char*, std::integral_constant<bool, StackEmpty_>, std::integral_constant<bool, AllowComments_>, bool) [with bool StackEmpty_ = true; bool IsKey_ = true; Handler = boost::json::detail::handler]: Assertion `*cs == '\x22'' failed.

This changes `sentinel()` to return the address of a static variable instead, which is guaranteed to be unique.

### DIFF
--- a/include/boost/json/basic_parser_impl.hpp
+++ b/include/boost/json/basic_parser_impl.hpp
@@ -217,8 +217,11 @@ const char*
 basic_parser<Handler>::
 sentinel()
 {
+    // the "+1" ensures that the returned pointer is unique even if
+    // the given input buffer borders on this object
     return reinterpret_cast<
-        const char*>(this);
+        const char*>(this) + 1;
+    return nullptr;
 }
 
 template<class Handler>
--- a/test/stream_parser.cpp
+++ b/test/stream_parser.cpp
@@ -17,6 +17,7 @@
 
 #include <sstream>
 #include <iostream>
+#include <cstring>
 
 #include "parse-vectors.hpp"
 #include "test.hpp"
@@ -1253,6 +1254,19 @@ R"xx({
         BOOST_TEST(serialize(t.jv) == "[]");
     }
 
+    // https://github.com/boostorg/json/pull/814
+    void
+    testSentinelOverlap()
+    {
+        struct {
+            char buffer[8];
+            boost::json::stream_parser p;
+        } s;
+        memcpy(s.buffer, "{\"12345\"", 8);
+        s.p.write(s.buffer, sizeof(s.buffer));
+        s.p.write(":0}", 3);
+    }
+
     //------------------------------------------------------
 
     void
@@ -1276,6 +1290,7 @@ R"xx({
         testDupeKeys();
         testIssue15();
         testIssue45();
+        testSentinelOverlap();
     }
 };
 


grisumbras:
This should work, but may affect performance. I wanted to see automatic benchmark results. @MaxKellermann also please add a test that's failing without your change (use one commit).

grisumbras:
Btw, alternative implementation would be to return the address of any parser's member except the first.

MaxKellermann:
> Btw, alternative implementation would be to return the address of any parser's member except the first.

That's fragile because it can break when you reorder the members. But you could add `+1` to the char-casted `this` pointer.

MaxKellermann:
I added a failing unit test and I changed the implementation to return `(char*)this + 1`; this was necessary because my first approach was still faulty - theoretically (though unlikely), this global variable could be the end of a global buffer, leading to the same crash as before.

grisumbras:
this + 1 will end up within the handler. So, I wonder if someone may try feeding the parser with something within the handler.

MaxKellermann:
> this + 1 will end up within the handler

True; but note that without this PR, it was already within the handler, so this PR is an improvement for sure.

What about using `nullptr` as sentinel? That would be cheaper than comparing pointers with `this` (i.e. with a CPU register). Are there corner cases where a valid non-sentinel value can be `nullptr`?

grisumbras:
Can you please squash the commits?

sdkrystian:
@vinniefalco this reason why `sentinel` returned `this` was because it would be already loaded in some register when the function was called, so it could simply be moved into `rax` via a register-to-register move (which don''t require any execution resources on modern CPUs). So while this fix may work, it may just be better to return `nullptr` (which I believe should work; possibly even better than the old implementation).

vinniefalco:
Interesting!!

MaxKellermann:
> it may just be better to return `nullptr` (which I believe should work; possibly even better than the old implementation).

[I had the same idea yesterday](https://github.com/boostorg/json/pull/814#issuecomment-1322337868), but

> This idea is so obvious that I thought there must be a reason why this wasn't already done

If you confirm there isn't any reason, I'll happily change this PR to `nullptr`!

vinniefalco:
try `nullptr`

vinniefalco:
Another possibility is to privately derive `basic_parser_impl` from a struct that has a single private char in it:
```
class detail::pad
{
  char c__ = 0;
};
```

MaxKellermann:
> Another possibility is to privately derive `basic_parser_impl` from a struct that has a single private char in it:

You'd need to have at least two chars, and return a pointer to the second one, to ensure that no preceding buffer can end at the returned pointer, which is the whole problem here.

MaxKellermann:
> Well no, because `sentinel()` is a member of `basic_parser_impl` so `this` would refer to the `basic_parser_impl` part and not the `detail::pad` part.

They have the same address in your example (most likely, but that may depend on the compiler-specific C++ ABI).

> one past `pad`

No. Let me godbolt this for you :-) https://godbolt.org/z/63T8c4vx8

vinniefalco:
:( That is rather unfortunate

MaxKellermann:
If you want it watertight, I can change this PR to add a `char sentinel[2]` field and return `&sentinel[1]`. Do you want that?
For my taste, the current state or the PR is "good enough", but it's not (strictly) watertight.
I'd love to have `nullptr`, but that may require fixing some existing quirks - if you want that, I can give that a try.
Let me know your preference.

grisumbras:
How would that be better than just returning an address to any member? Or this + 1?

MaxKellermann:
> How would that be better than just returning an address to any member?

If you return the address of any member, it's only watertight if you can guarantee that this member is not directly preceded by a buffer to be parsed. You can look at the current layout of the class and speculate it's not, but that's fragile; if eventually the code gets refactored and members get reordered (which is a perfectly reasonable thing to do), this assumption may no longer be true and the code may break.

Having a `char sentinel[2]` and using `&sentinel[1]` guarantees that the pointer is never preceded by a buffer, because `&sentinel[1]` is guaranteed to be preceded by `&sentinel[0]`, no matter which C++ ABI, and no matter how you reorder fields. This is the strictly watertight solution.

> Or this + 1?

It was you who wrote: "this + 1 will end up within the handler. So, I wonder if someone may try feeding the parser with something within the handler."
It's not whatertight. It's only safe against parsing buffers outside of the parser object, but not against buffers which happen to be inside.

vinniefalco:
`(reinterpret_cast<char const*>(this) + 1)`

include/boost/json/basic_parser_impl.hpp
```diff
@@ -217,8 +217,10 @@ const char*
 basic_parser<Handler>::
 sentinel()
 {
+    // the "+1" ensures that the returned pointer is unique even if
+    // the given input buffer borders on this object
     return reinterpret_cast<
```
WPMGPRoSToTeMa, line 222: `static_assert(sizeof(*this) > 1);`

MaxKellermann: `class basic_parser` has lots of fields, and is always larger than 1 byte, unless this class gets refactored and all fields get removed, which isn't going to happen. I don't think this `static_assert` adds real value, it's trivial to see it's always true.

vinniefalco: Agreed, lol

grisumbras:
The worst part of making `sentinel()` return `nullptr` is that it will put a precondition on `write_some` (you wouldn't be able to do `parser.write_some(true, nullptr, 0, ec)`. This isn't necessarily bad, but when GCC started optimising around a similar precondition in `memcpy`, it broke a lot of code. 

I would stick to returning a pointer into parser, even if that means some reduction in performance. I also don't like adding unnecessary fields just for this purpose, I would prefer returning a pointer to an existing field. We already keep layout of `value` et al. fixed, doing something similar for `basic_parser` shouldn't be a problem.

On the other hand, I'm not necessarily against returning `this + 1`, but we might want to document the fact that people shouldn't feed to the parser buffers inside `handler`.

@vinniefalco your thoughts?

grisumbras:
@MaxKellermann can you please rebase on current develop and re-push. I will merge it then.

MaxKellermann:
> @MaxKellermann can you please rebase on current develop and re-push. I will merge it then.

done

grisumbras:
Thank you for your contribution.

PR #819, opened 2022-12-06
Feature/context conversions

### LINKED ISSUE #647: [feature request] Domain specific conversions
I mentioned that to Vinnie already, but I'd profit from a context object for json, and after some thinking I'd just go for an additional parameter getting passed by const ref and propagated to all the tag_invokes. I don't know if the current overload resolution would work with this though.

```cpp
json::value_from(T, some_tag{});

void
tag_invoke(json::value_from_tag, json::value &target, const U &src, const some_tag &);

json::value_to<T>(json::value, some_tag{});

void
tag_invoke(json::value_to_tag<U>, json::value &target, const some_tag &);
```

but that would allow me to attach a context, e.g. to turn all the std::chrono types into ISO8601 string including proper adl for that:

```cpp
namespace my_date_library
{
struct my_date_tag;

void
tag_invoke(json::value_from_tag, json::value &target, const U &src, const my_date_tag &);

void
tag_invoke(json::value_to_tag<U>, json::value &target, const my_date_tag &);

}
```

and then I can combine those into my own tag:

```cpp
struct klemens_tag: my_date_library::my_date_tag {};
```

### LINKED ISSUE #854: value_to object with stateful allocator such as std::pmr::string
How can we construct a `std::pmr::string` using `value_to`, with user-provided `std::pmr::memory_resource*`?
Here, the object and the json value are using different `memory_resource` sources.

The customization function's signature is `T tag_invoke( const value_to_tag< T >&, const value& )`, which does not take any `Allocator` as parameter. So there is no way to provide the allocator.

It seems we can only construct `std::pmr::string` from `std::pmr::get_default_resource()`?

### LINKED ISSUE #871: Is there a way to customize conversion for a type outside of its namespace?
### Version of Boost

1.81

### Question

I want to make `value_to` and `value_from` conversions for a type `T` defined in third-party library. It works properly if `tag_invoke` function is defined in `boost::json` namespace or the same namespace as `T`, but it causes compile error if the function is defined in other namespaces.

What I would like to do is as the follows.

```C++
// third_party.hpp

namespace third_party {
    class TheirType;
}
```

```C++
// my_app.cpp

#include "third_party.hpp"
#include <boost/json.hpp>

using third_party::TheirType;

namespace {  
    TheirType tag_invoke(const boost::json::value_to_tag<TheirType>, const boost::json::value& json) {
        /* implementation*/
    }

    void tag_invoke(const boost::json::value_from_tag&, boost::json::value& json, const TheirType& theirs) {
        /* implementation*/
    }
}

namespace my_app {
    void my_func(const boost::json::value& json) {
        auto t = boost::json::value_to<TheirType>(json);
    }
}
```

If possible, I want to define `tag_invoke` overload function in unnamed namespace or `my_app` namespace not to add changes for `third_party` and `boost::json` namespaces. Is there a way to do it?

Let me add that it is not very important because I can work around it by defining it in `boost::json` namespace.

grisumbras:
Fix #647
Fix #854
Fix #871

### DIFF
--- a/include/boost/json/conversion.hpp
+++ b/include/boost/json/conversion.hpp
@@ -11,11 +11,19 @@
 #define BOOST_JSON_CONVERSION_HPP
 
 #include <boost/json/detail/config.hpp>
+#include <boost/json/fwd.hpp>
 
 #include <type_traits>
 
 namespace boost {
 namespace json {
+
+namespace detail {
+
+template< class Ctx, class T, class Dir >
+struct supported_context;
+
+} // namespace detail
 
 /** Customization point tag.
 
--- a/doc/qbk/conversion/context.qbk
+++ b/doc/qbk/conversion/context.qbk
@@ -1,0 +1,123 @@
+[/
+    Copyright (c) 2022 Dmitry Arkhipov (grisumbras@yandex.ru)
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+    Official repository: https://github.com/cppalliance/json
+]
+
+[/-----------------------------------------------------------------------------]
+
+
+[section Contextual conversions]
+
+Previously in this section we've been assuming that there is a particular
+fitting JSON representation for a type. But this is not always the case. Often
+one needs to represent particular value with JSON of certain format in one
+situation and with another format in a different situation. This can be
+achieved with Boost.JSON by providing an extra argument\U00002014context.
+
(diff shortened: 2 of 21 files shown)


include/boost/json/conversion.hpp
```diff
@@ -322,6 +322,9 @@ struct is_described_class;
 template<class T>
 struct is_described_enum;
```
vinniefalco, line 323: is this header included in `<boost/json.hpp>`?

grisumbras: Yes

PR #849, opened 2023-01-28
use alignment facilities from Core and Align

grisumbras:


vinniefalco:
This is very good, yes it belongs in its own commit

PR #855, opened 2023-02-08
Skip bench if dependencies are missing.

grafikrobot:
If one exec b2 at the json-root without running the clone.sh script one will get header-not-found errors. This change prevents the bench from building when the required extra source files are not present.

### DIFF
--- a/bench/Jamfile
+++ b/bench/Jamfile
@@ -7,9 +7,13 @@
 # Official repository: https://github.com/boostorg/json
 #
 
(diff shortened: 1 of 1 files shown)


grisumbras:
What about just making the target explicit?

grisumbras:
Or removing `build-project bench ;` from the root Jamfile?

grisumbras:
I've merged alternative fix in #857

PR #863, opened 2023-03-01
non-allocating category messages

### LINKED ISSUE #862: error_category best practices
Each library-defined error_category should have these members as written:

        std::string
        message( int ev ) const override
        {
            return text(ev);
        }

        char const*
        message(
            int ev,
            char *dest,
            std::size_t len) const noexcept
        {
            if(len > 0)
            {
                auto s = text(ev);
                auto n = s.size();
                if( n >= len)
                    n = len - 1;
                std::memcpy(
                    dest, s.data(), n);
                dest[n] = 0;
            }
            return dest;
        }
    };
Add string_view text( int ) const noexcept and put the message strings there.

grisumbras:
Fix #862

### DIFF
--- a/include/boost/json/impl/error.ipp
+++ b/include/boost/json/impl/error.ipp
@@ -26,8 +26,8 @@ make_error_code(error e)
             return "boost.json";
         }
 
-        std::string
-        message(int ev) const override
+        char const* message( int ev, char*, std::size_t ) const noexcept
+            override
         {
             switch(static_cast<error>(ev))
             {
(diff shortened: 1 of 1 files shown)


vinniefalco:
`make_error_code` needs to be inline and visible, like this: 
https://github.com/vinniefalco/http_proto/blob/871c20cc38e337880363a5efa2649a22806e1bd0/include/boost/http_proto/impl/error.hpp#L85

grisumbras:
Why inline?

vinniefalco:
performance

PR #865, opened 2023-03-01
inline make_error_code

grisumbras:


include/boost/json/error.hpp
```diff
@@ -157,6 +157,47 @@ enum class condition
...
+    @par Exception Safety
+    No-throw guarantee.
+
+    @returns @ref error_code object corresponding to the value of `e`.
+
+    @param e The @ref error code to convert.
+*/
+constexpr
+error_code
+make_error_code(error e) noexcept;
```
vinniefalco, line 178: Why is this public now?

grisumbras:
@pdimov can you look at this? Does this look OK? I've looked at the change in CE (https://godbolt.org/z/qr443czfT) and it does seem to result in less code.

pdimov:
This looks OK to me, except that I don't think the extern declarations need to be marked as constinit.

grisumbras:
Are you saying that I should remove constinit or that I should have either constinit or extern?

Issue #599, opened 2021-08-03
Loss of precision during serialization/parsing plain double ?


marc-viala:
I'm starting to implement some serialization tools with Boost.Json but I've got some precision issues with double values. To demonstrate this "potential" issue, you will find hereafter a code snippet to reproduce it:

```
#include <boost/json.hpp>

BOOST_AUTO_TEST_CASE(bjson_double_serialize_ut)
{
  namespace bjs = boost::json;
  static double ref_values[] = {
      1217.2772861138403
    , -161.68713249779881
    , 267.04251495962637};

  std::string s;

  {
    const bjs::value jv = {
      ref_values[0] , ref_values[1], ref_values[2], 0.0, 0.0, 0.0};
    s = serialize(jv);
  }

  double values[3] = {};
  {
    const auto jv{bjs::parse(s)};
    values[0] = jv.as_array().at(0).as_double();
    values[1] = jv.as_array().at(1).as_double();
    values[2] = jv.as_array().at(2).as_double();
  }

  static constexpr auto eps{std::numeric_limits<double>::epsilon()};
  BOOST_TEST(std::abs(ref_values[0] - values[0]) <= eps);
  BOOST_TEST(std::abs(ref_values[1] - values[1]) <= eps);
  BOOST_TEST(std::abs(ref_values[2] - values[2]) <= eps);
}
```

When I compiling and running this code w/ Boost 1.75 and Microsoft Visual C++ version 14.2 (VS 2019) / Windows 10, I've got the following output:

> "bjson_double_serialize_ut": check std::abs(ref_values[0] - values[0]) <= eps has failed [2.2737367544323206e-13 > 2.2204460492503131e-16]
> "bjson_double_serialize_ut": check std::abs(ref_values[2] - values[2]) <= eps has failed [5.6843418860808015e-14 > 2.2204460492503131e-16]

From my point of view, I'm expecting to get a tolerance equal or inf. to my "double machine epsilon". Is this correct and how can I address it?

Marc

marc-viala:
After some iterations, the first version of my unit test passes w/ Boost 1.76 and fails w/ 1.75. But maybe the issue of loss of precision during serialization/parsing is not fully addressed w/ Boost 1.76 because this new unit test hereafter (more exhaustive) is failing w/ Boost 1.76:

```
BOOST_AUTO_TEST_CASE(bjson_double_serialize_ut)
{
  static const std::vector ref_values = {
      -0.27006296145688152
    , -0.42448451824686562
[...]

PR #890, opened 2023-05-17
Precise mode for parsing numbers

grisumbras:
This is technically a fix to #599, but we want a better implementation than `strtod` (because it is very slow). Closer to the release I'll take some files from Charconv.

### DIFF
--- a/bench/bench.cpp
+++ b/bench/bench.cpp
@@ -53,6 +53,7 @@ using clock_type = std::chrono::steady_clock;
 
 ::test_suite::debug_stream dout(std::cerr);
 std::stringstream strout;
+parse_options popts;
 
 #if defined(__clang__)
 string_view toolset = "clang";
@@ -276,7 +277,7 @@ public:
         string_view s,
         std::size_t repeat) const override
     {
-        stream_parser p;
+        stream_parser p({}, popts);
         while(repeat--)
         {
             p.reset();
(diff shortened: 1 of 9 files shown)


include/fast_float/parse_number.h
```diff
@@ -0,0 +1,230 @@
+#ifndef FASTFLOAT_PARSE_NUMBER_H
+#define FASTFLOAT_PARSE_NUMBER_H
+
```
vinniefalco, line 3: You need copyrights and please "hpp" extension not "h"

include/fast_float/parse_number.h
```diff
@@ -0,0 +1,230 @@
...
+#include "ascii_number.h"
+#include "decimal_to_binary.h"
+#include "digit_comparison.h"
```
vinniefalco, line 6: These need to be standard full boost paths, e.g. <boost/json/ umm... wait a minute you have this in <boost/fastfloat/> ? That's not allowed

include/fast_float/parse_number.h
```diff
@@ -0,0 +1,230 @@
...
+#include "float_common.h"
+
+#include <cmath>
+#include <cstring>
+#include <limits>
+#include <system_error>
+
+namespace fast_float {
```
vinniefalco, line 14: Where's the boost namespace?

include/fast_float/parse_number.h
grisumbras: Turns out fast_float isn't permissive enough, so I had to remove it anyway.

PR #891, opened 2023-05-18
refactor handling of literals in parser

grisumbras:
This PR will be particularly relevant, when we add handling of non-standard literals Infinity, -Infinity, and NaN.


grisumbras:
Moved to #894

PR #892, opened 2023-05-19
accept numbers with large exponent

grisumbras:
This changes treatment of numbers with exponents larger than `INT_MAX`. The current behaviour is somewhat arbitrary: we reject numbers with exponents that do not fit into `INT_MAX`, but we accept numbers with exponents that do not fit in `double` (we parse those numbers as infinity).

This PR removes the inconsistency by just setting exponent to `INT_MAX`. The alternative solution to inconsistency -- making it an error if the exponent doesn't fit into `double` is undesirable. Parsing such numbers as infinity/zero is actually standard practice.


pdimov:
This will theoretically fail for numbers like `0.000....0001e2147483648` where the number of zeroes is close to 2147483647, but I suppose this can never happen in practice.

Does `TEST_GOOD` check that the parsed value is zero or infinity in these cases?

grisumbras:
> Does `TEST_GOOD` check that the parsed value is zero or infinity in these cases?

No, it only tests that parse doesn't result in an error.

grisumbras:
> This will theoretically fail for numbers like `0.000....0001e2147483648` where the number of zeroes is close to 2147483647, but I suppose this can never happen in practice.

Maybe we should consider such numbers errors instead (as far as I can see, we don't even check for overflow of the corresponding variable at the moment).

pdimov:
Considering them errors should be fine.

PR #897, opened 2023-05-26
constexpr handler member functions

grisumbras:


grisumbras:
@pdimov do you think there's any utility to this change? It seems that compilers can see that the handler's member functions never return false any way.

pdimov:
I wouldn't expect it to make any difference.

PR #898, opened 2023-05-26
Add library-detail.adoc for website contents

alandefreitas:


grisumbras:
Please, in the future don't merge branches yourself

grisumbras:
Also, I've noticed that the links you've copied from README are wrong (they should direct users to boostorg GH and Boost website instead).

PR #905, opened 2023-06-08
use Charconv for precise parsing

grisumbras:


include/boost/json/detail/charconv/detail/config.hpp
```diff
@@ -0,0 +1,57 @@
...
+#if defined(BOOST_HAS_INT128)
+#  define BOOST_JSON_INT128_MAX  (boost::int128_type)(((boost::uint128_type) 1 << 127) - 1)
+#  define BOOST_JSON_UINT128_MAX ((2 * (boost::uint128_type) BOOST_JSON_INT128_MAX) + 1)
+#endif
+
+#ifndef BOOST_NO_CXX14_CONSTEXPR
+#  define BOOST_JSON_CXX14_CONSTEXPR BOOST_CXX14_CONSTEXPR
+#else
+#  define BOOST_JSON_CXX14_CONSTEXPR inline
+#endif
```
mborland, line 22:
```suggestion
#ifndef BOOST_NO_CXX14_CONSTEXPR
#  define BOOST_JSON_CXX14_CONSTEXPR BOOST_CXX14_CONSTEXPR
#  define BOOST_JSON_CXX14_CONSTEXPR_NO_INLINE BOOST_CXX14_CONSTEXPR
#else
#  define BOOST_JSON_CXX14_CONSTEXPR inline
#  define BOOST_JSON_CXX14_CONSTEXPR_NO_INLINE
#endif
```

You'll need the no inline version in fast float because I think everything marked `CXX14_CONSTEXPR` is also `BOOST_FORCEINLINE`. They'll collide with C++11

grisumbras: Indeed

PR #907, opened 2023-06-10
Fix object strong guarantee

### LINKED ISSUE #763: object operator[] and insert functions invalidate iterators/references

### LINKED ISSUE #906: Operator[] / emplace / insert_or_assign on existing key may cause iter invalidation
### Version of Boost

1.82

https://godbolt.org/z/MeKzYbqsv

Crashes, because `operator []` called with existing key as an argument causes reallocation and iterator invalidation ignoring the fact the key hasn't really been inserted. According to docs, iterators are not affected if the key isn't inserted.

LLVM Clang 16.0.0
Originally found on MacOS Ventura 13.2.1

#### Notes 
Relevant documentation for operator[] ([reference](https://www.boost.org/doc/libs/master/libs/json/doc/html/json/ref/boost__json__object/operator_lb__rb_.html)) declares that iterator invalidation happens only on successful value insertion. However, [as it can be found in source code](https://github.com/boostorg/json/blob/ee01580cfde818e31d8ea41c2e4056bcff9e3018/include/boost/json/impl/object.ipp#L598) for boost 1.82 (older versions too, probably), operator[] works via calling `emplace(key, nullptr)`. 

The real "problem" lies in the `boost::json::object::emplace` function. It is said in the docs it works the same way as `[]` - invalidation only upon insertion. Unfortunately, it's not stated that `emplace` always ([ref to boost 1.82 source](https://github.com/boostorg/json/blob/ee01580cfde818e31d8ea41c2e4056bcff9e3018/include/boost/json/impl/object.hpp#L410)) reserves memory for at least `size() + 1` elements before checking should the element be inserted or not - so iterators **could** be invalidated (if you are unlucky) even on unsuccessful insertion.

#### Suggested change

Haven't prepared the MR though, but I suppose it would be enough to move lines with memory reservation in `boost::json::object::emplace` and `boost::json::object::insert_or_assign` below checking for existing with (maybe) some extra checks (as `detail::find_in_object` asserts for non-zero capacity, for example).

grisumbras:
Fix #763

### DIFF
--- a/include/boost/json/impl/object.hpp
+++ b/include/boost/json/impl/object.hpp
@@ -154,6 +154,7 @@ public:
 class object::revert_insert
 {
     object* obj_;
+    table* t_ = nullptr;
     std::size_t size_;
 
     BOOST_JSON_DECL
@@ -163,24 +164,38 @@ class object::revert_insert
 public:
     explicit
     revert_insert(
-        object& obj) noexcept
+        object& obj,
+        std::size_t capacity)
         : obj_(&obj)
         , size_(obj_->size())
     {
+        if( capacity > obj_->capacity() )
+            t_ = obj_->reserve_impl(capacity);
     }
 
     ~revert_insert()
     {
         if(! obj_)
             return;
+
         destroy();
-        obj_->t_->size = static_cast<
-            index_t>(size_);
+        if( t_ )
+        {
+            table::deallocate( obj_->t_, obj_->sp_ );
+            obj_->t_ = t_;
+        }
+        else
+        {
+            obj_->t_->size = static_cast<index_t>(size_);
+        }
     }
 
     void
     commit() noexcept
     {
+        BOOST_ASSERT(obj_);
+        if( t_ )
+            table::deallocate( t_, obj_->sp_ );
         obj_ = nullptr;
     }
 };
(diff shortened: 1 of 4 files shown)


vinniefalco:
There is no requirement that capacity is restored to the previous value on an exception.. it is ok to increase capacity and then later fail the insertion - and the object is returned to the caller with the increase capacity.

grisumbras:
If new storage is allocated, then iterators are invalidated and thus, the function doesn't provide the string guarantee.

grisumbras:
It's not too hard to revert storage change in our case, so I don't see why we shouldn't. Strong guarantee is a good property.

PR #912, opened 2023-06-29
Fix type of pointer to _umul128

mborland:
`_umul128` expects a pointer of type `unsigned long long*` instead of `uint64_t*`.


grisumbras:
Thanks for the patch!

PR #915, opened 2023-07-12
Use Boost.Endian for endianness

grisumbras:


vinniefalco:
Refresh my memory - what does JSON have to do with endianness at all?

grisumbras:
At several points we load buffers into numbers for bitwise comparison. This seems to have been done for optimisation puproses

pdimov:
Instead of reversing in place, you can just use the Endian load/store functions.

PR #921, opened 2023-07-21
fix inconsistent choice of init list constructor

### LINKED ISSUE #920: value constructor behaviour inconsistent across compilers (initializer_list)
- Boost 1.82.0

Repro:

```c++
#include <boost/json/src.hpp>
namespace json = boost::json;

int main() {
    json::value a, b{a};
    assert(b.is_array()); // GCC pass, Clang fail
}
```

Compare results [On Compiler Explorer](https://godbolt.org/z/oYv5KKvsT)

I'm not qualified to know which compiler is in error (I'm tempted to side with GCC, but I'm not a language-lawyer), but no doubt someone else knows. (Related context: https://stackoverflow.com/q/76724646/85371)

grisumbras:
On some compilers `value{x}` invokes `initializer_list` constructor, on others it is equivalent to `value(x)`. This is very problematic, but this isn't something we can fix. On the other hand, we _can_ make init list construction to be equivalent to `value(x)`, if the size of init list is one. This commit does exactly that.

Fix #920

### DIFF
--- a/include/boost/json/impl/value.ipp
+++ b/include/boost/json/impl/value.ipp
@@ -231,13 +231,28 @@ value(
     storage_ptr sp)
 {
     if(value_ref::maybe_object(init))
+    {
         ::new(&obj_) object(
             value_ref::make_object(
                 init, std::move(sp)));
+    }
     else
-        ::new(&arr_) array(
-            value_ref::make_array(
-                init, std::move(sp)));
+    {
+#ifndef BOOST_JSON_LEGACY_INIT_LIST_BEHAVIOR
+        if( init.size() == 1 )
+        {
+            ::new(&sca_) scalar();
+            value temp = init.begin()->make_value( std::move(sp) );
+            swap(temp);
+        }
+        else
+#endif
+        {
+            ::new(&arr_) array(
+                value_ref::make_array(
+                    init, std::move(sp)));
+        }
+    }
 }
 
 //----------------------------------------------------------
--- a/include/boost/json/value.hpp
+++ b/include/boost/json/value.hpp
@@ -1014,11 +1014,16 @@ public:
 
     /** Construct from an initializer-list
 
-        If the initializer list consists of key/value
-        pairs, an @ref object is created. Otherwise
-        an @ref array is created. The contents of the
-        initializer list are copied to the newly constructed
-        value using the specified memory resource.
+        @li If the initializer list consists of key/value
+        pairs, an @ref object is created; otherwise,
+
+        @li if the size of the initializer list is exactly 1, the object is
+        constructed directly from that sole element; otherwise,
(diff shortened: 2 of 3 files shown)


vinniefalco:
Hey good idea !

nkolotov:
¿What if caller actually wants to produce an array with a single item? Some people may already be using code that supplies `initializer_list` explicitly expecting array to be produced.
```
void Pack(::std::initializer_list<::boost::json::value_ref> items)
{
    ::boost::json::value json{items};
    ::boost::json::array & array{json.as_array()};
    // do something with array...
}
```
And proposed change will break this code when only a single item is supplied.

Also i'm not sure about `::new(this) value (...`. ¿Is an attempt to reuse storage of the object which lifetime hasn't started yet a well-defined behavior? ¿Will access to newly created object through reference to the old one be valid?

grisumbras:
The proposed change is the only way I could come up with that would make the behaviour consistent. The problem of behaviour being inconsistent between compilers have been haunting us for years.

That being said, creating an array from one element init list is very easy:
```
std::initializer_list<value_ref> items = {x};
value jv = array(items);
``` 
or
```
value jv = array{x};
```

grisumbras:
> Also i'm not sure about `::new(this) value (...`. ¿Is an attempt to reuse storage of the object which lifetime hasn't started yet a well-defined behavior? ¿Will access to newly created object through reference to the old one be valid?

How could there be a reference to an object that does not exist yet?

nkolotov:
> The proposed change is the only way I could come up with that would make the behaviour consistent.
However you make behavior of constructor taking `initializer_list` inconsistent. Potentially breaking existing code.

> How could there be a reference to an object that does not exist yet?
In case if `::new(this)`, `this` is a pointer to the object being currently constructed. And if object here turns out to be not transparently replaceable then `this` pointer as well as any other names referring to the current object will become invalid. Also even if object *is* transparently replaceable, `::new(this)` is dangerous in other ways since `::boost::json::value` is not trivially destructible type.

grisumbras:
> However you make behavior of constructor taking `initializer_list` inconsistent. Potentially breaking existing code.

No, it will not make it inconsistent. It will be consistent with documented behaviour (which reminds me that I should have amended documentation). And yes, it will break existing code. So, I'm leaning towards temporary providing a macro to disable the new behaviour. 

> In case if `::new(this)`, `this` is a pointer to the object being currently constructed. And if object here turns out to be not transparently replaceable then `this` pointer as well as any other names referring to the current object will become invalid. Also even if object _is_ transparently replaceable, `::new(this)` is dangerous in other ways since `::boost::json::value` is not trivially destructible type.

Ok, just to be sure I will replace `new(this)` with a different implementation.

Issue #927, opened 2023-08-11
__emulu() is not available on arm


liubing:
### Version of Boost

boost 1.83beta

### Steps necessary to reproduce the problem

Compile the library on arm64 / windows platform.

### All relevant compiler information
Error message:
```
D:\Projects\pts\pts\include\boost\json\detail\charconv\detail\emulated128.hpp(62): error C3861: '__emulu': identifier not found
```
It can be fixed changing  
```
static inline std::uint64_t umul64(std::uint32_t x, std::uint32_t y) noexcept
{
#if defined(BOOST_JSON_HAS_MSVC_32BIT_INTRINSICS)
    return __emulu(x, y);
#else
    return x * static_cast<std::uint64_t>(y);
#endif
}
```
to
```
static inline std::uint64_t umul64(std::uint32_t x, std::uint32_t y) noexcept
{
#if defined(BOOST_JSON_HAS_MSVC_32BIT_INTRINSICS) && !defined(_M_ARM)
    return __emulu(x, y);
#else
    return x * static_cast<std::uint64_t>(y);
#endif
}
```

liubing:
#928 doesn't really fix this problem. It should be
```
#if defined(BOOST_JSON_HAS_MSVC_32BIT_INTRINSICS) && !defined(_M_ARM)
```
not
```
#if defined(BOOST_JSON_HAS_MSVC_32BIT_INTRINSICS) && !defined(_M_ARM64)
```

mborland:
That's my mistake. The PR has been updated.

liubing:
Thank you.

Issue #926, opened 2023-08-11
_umul128() is not available on arm64


liubing:
### Version of Boost

boost 1.83beta

### Steps necessary to reproduce the problem

Compile the library on arm64 / windows platform.

### All relevant compiler information
Error message:
```
D:\Projects\pts\pts\include\boost\json\detail\charconv\detail\emulated128.hpp(79): error C3861: '_umul128': identifier not found
```
It can be fixed changing  
```
    #elif defined(BOOST_JSON_HAS_MSVC_64BIT_INTRINSICS)

    std::uint64_t high;
    std::uint64_t low = _umul128(x, y, &high);
    return {high, low};

    // https://developer.arm.com/documentation/dui0802/a/A64-General-Instructions/UMULH
    #elif defined(__arm__)
```
to
```
    #elif defined(BOOST_JSON_HAS_MSVC_64BIT_INTRINSICS) && !defined(_M_ARM64)

    std::uint64_t high;
    std::uint64_t low = _umul128(x, y, &high);
    return {high, low};

    // https://developer.arm.com/documentation/dui0802/a/A64-General-Instructions/UMULH
    #elif defined(__arm__)

```

PR #928, opened 2023-08-11
Fix use of intrinsics on windows ARM platforms

mborland:
Closes: #926 
Closes: #927 

Changes use of intrinsics: https://learn.microsoft.com/en-us/cpp/intrinsics/umul128?view=msvc-170 and https://learn.microsoft.com/en-us/cpp/intrinsics/emul-emulu?view=msvc-170

### DIFF
--- a/include/boost/json/detail/charconv/detail/emulated128.hpp
+++ b/include/boost/json/detail/charconv/detail/emulated128.hpp
@@ -58,7 +58,7 @@ struct uint128
 
 static inline std::uint64_t umul64(std::uint32_t x, std::uint32_t y) noexcept
 {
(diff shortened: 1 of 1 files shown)


grisumbras:
I feel like rather than doing `if defined(BOOST_JSON_HAS_MSVC_64BIT_INTRINSICS) && !defined(_M_ARM64)` and similar, we should have `BOOST_JSON_HAS_MSVC_X86_64_INTRINSICS`, `BOOST_JSON_HAS_MSVC_ARM64_INTRINSICS` `BOOST_JSON_HAS_MSVC_X86_INTRINSICS`, `BOOST_JSON_HAS_MSVC_ARM32_INTRINSICS`. Or does the overlap make it impractical?

grisumbras:
Nah, let's merge it as-is. I'm expecting to remove it with dependency on Boost.Charconv soon anyway.

PR #946, opened 2023-10-02
Specify output location of bench executable

sdarwin:
This will build the benchmark executable as` ./bin.v2/bench` instead of `bin.v2/libs/json/bench/clang-linux-14/release/cxxstd-20-iso/link-static/bench`. By outputting the exe in the same place each time (similar to how b2 generates it's own file as `./b2`) it is easier to find and won't change later.  

It would be more convenient if the output locations were ` ./bin.v2/clang/bench` and ` ./bin.v2/gcc/bench`.  That requires knowing the 'toolset'.  Interestingly from the faqs, "How do I get the current value of feature in Jamfile? This is not possible, since Jamfile does not have "current" value of any feature, be it toolset, build variant or anything else. " $(toolset) doesn't have a value.  We can't define the output location to be  ` ./bin.v2/$(toolset)/bench`

After merging this into develop branch `runbenchmarks.py` will break.   I will need to modify the script, manually copy `bench` into ` ./bin.v2/clang/bench` and ` ./bin.v2/gcc/bench`, and run benchmarks from there.  @grisumbras if it looks ok, go ahead and merge.

### DIFF
--- a/bench/Jamfile
+++ b/bench/Jamfile
@@ -28,4 +28,5 @@ exe bench :
     <include>../test
     $(has_nlohmann_json)<define>BOOST_JSON_HAS_NLOHMANN_JSON
     $(has_rapidjson)<define>BOOST_JSON_HAS_RAPIDJSON
+    : <location>.//bin.v2
     ;


sdarwin:
https://benchmark.cppalliance.org/ processes commits going back to 2020 in order to compare them. This pull request won't be able to retroactively modify earlier copies of the bench Jamfile. So [runbenchmarks.py](https://github.com/cppalliance/Infrastructure-Docs/blob/master/benchmarks/json/runbenchmarks.py) will maintain it's own copy of the file, and write it to disk each time.  That's in place now.

grisumbras:
Your change will break any multiconfiguration build that builds project `json/bench`, because several targets would try to create the same file.

In #947 I instead added a different target that has to be requested explicitly. The target is `bench-local`. You can get it built with e.g.

```
b2 toolset=gcc bench//bench-local
```

Note that for example
```
b2 toolset=gcc,clang
b2 toolset=gcc bench//bench-local
bench/bench-local bench/data/*
b2 toolset=clang bench//bench-local
bench/bench-local bench/data/*
```
**will not** work as expected, because `b2 toolset=gcc,clang` builds `bench//bench` for both gcc and clang, then `b2 toolset=gcc bench//bench-local` builds `bench//bench-local` for gcc, and then `b2 toolset=clang bench//bench-local` checks if `bench//bench-local` needs to be updated, sees that it's more recent than its dependency (`bench//bench`) and decides not to update it.

If this is a problem, I can set `bench//bench-local` to always be considered out of date.

sdarwin:
Superseded by the pull request "Targets for benchmarks". Closing.

PR #947, opened 2023-10-04
Targets for benchmarks

grisumbras:


grisumbras:
@sdarwin is this PR good enough for your CI changes?

sdarwin:
Yes, thanks. I will copy this new jamfile into `runbenchmarks.py`.

PR #948, opened 2023-10-11
Feature/stack nontrivials

grisumbras:


include/boost/json/detail/impl/stack.hpp
```diff
@@ -0,0 +1,176 @@
...
+    the lo end of the buffer and the
+    trivial objects at the hi end:
+
+  base_        {free}     base_+cap_
+    |<------>|<------>|<------>|
+      size0_            size1_
+*/
+//--------------------------------------
+
+struct stack::non_trivial
```
vinniefalco, line 32: This looks awfully familiar, didn't I already do something similar?

Issue #968, opened 2024-01-05
Linux gcc compilation failures with --pedantic-errors flag: error: extra ‘;’


nigels-com:
### Version of Boost

1.84.0

### Steps necessary to reproduce the problem

Ubuntu 22.04, g++ 11.4.0

### Example

`--pedantic-errors` worked fine for Boost 1.83.0 boost::json, and previous versions of Boost.

It would be be nice to have this fixed for a possible 1.83.1, this is a blocker for us,  `--pedantic-errors` mode is strongly encouraged internally.  (Rightly or wrongly)

```
$ g++ -c validate.cpp -I$(pwd)/../include -I$(pwd)/../tmp/boost_1_84_0 --pedantic-errors
In file included from /home/nigels/dev/boost-json/example/../include/boost/json/parse_into.hpp:19,
                 from /home/nigels/dev/boost-json/example/../include/boost/json.hpp:26,
                 from validate.cpp:16:
/home/nigels/dev/boost-json/example/../include/boost/json/detail/parse_into.hpp:870:47: error: extra ‘;’ [-Wpedantic]
  870 |     BOOST_JSON_HANDLE_EVENT( on_object_begin );
      |                                               ^
      |                                               -
/home/nigels/dev/boost-json/example/../include/boost/json/detail/parse_into.hpp:871:45: error: extra ‘;’ [-Wpedantic]
  871 |     BOOST_JSON_HANDLE_EVENT( on_object_end );
      |                                             ^
      |                                             -
/home/nigels/dev/boost-json/example/../include/boost/json/detail/parse_into.hpp:882:10: error: extra ‘;’ [-Wpedantic]
  882 |         };
      |          ^
      |          -
/home/nigels/dev/boost-json/example/../include/boost/json/detail/parse_into.hpp:913:10: error: extra ‘;’ [-Wpedantic]
  913 |         };
      |          ^
      |          -
/home/nigels/dev/boost-json/example/../include/boost/json/detail/parse_into.hpp:929:43: error: extra ‘;’ [-Wpedantic]
  929 |     BOOST_JSON_HANDLE_EVENT( on_key_part );
      |                                           ^
      |                                           -
/home/nigels/dev/boost-json/example/../include/boost/json/detail/parse_into.hpp:930:38: error: extra ‘;’ [-Wpedantic]
  930 |     BOOST_JSON_HANDLE_EVENT( on_key );
      |                                      ^
      |                                      -
[...]

PR #969, opened 2024-01-05
Fix for Issue #968 - extra ; in gcc --pedantic-errors mode

nigels-com:
Without this fix `g++ --pedantic-errors` will fail with errors, as discussed in Issue #968

Seems like a boost 1.84.0 regression, not seen with previous versions of boost.

```
nigels@zooropa:~/dev/boost-json/example$ g++ -c validate.cpp -I$(pwd)/../include -I$(pwd)/../tmp/boost_1_84_0 --pedantic-errors
nigels@zooropa:~/dev/boost-json/example$
```

### DIFF
--- a/include/boost/json/detail/parse_into.hpp
+++ b/include/boost/json/detail/parse_into.hpp
@@ -867,8 +867,8 @@ public:
                 F{ std::forward_as_tuple(ec, args...) } } ); \
     }
 
-    BOOST_JSON_HANDLE_EVENT( on_object_begin );
-    BOOST_JSON_HANDLE_EVENT( on_object_end );
+    BOOST_JSON_HANDLE_EVENT( on_object_begin )
+    BOOST_JSON_HANDLE_EVENT( on_object_end )
 
     struct do_on_array_begin
     {
@@ -879,7 +879,7 @@ public:
         bool operator()( I ) const
         {
             return get<I::value>(handlers).on_array_begin(ec);
-        };
+        }
     };
     bool on_array_begin( error_code& ec )
     {
@@ -910,7 +910,7 @@ public:
         bool operator()( I ) const
         {
             return get<I::value>(handlers).on_array_end(ec);
-        };
+        }
     };
     bool on_array_end( error_code& ec )
     {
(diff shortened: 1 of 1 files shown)


include/boost/json/detail/parse_into.hpp
```diff
@@ -867,8 +867,8 @@ class converting_handler<tuple_conversion_tag, T, P>
                 F{ std::forward_as_tuple(ec, args...) } } ); \
     }
 
-    BOOST_JSON_HANDLE_EVENT( on_object_begin );
-    BOOST_JSON_HANDLE_EVENT( on_object_end );
+    BOOST_JSON_HANDLE_EVENT( on_object_begin )
```
grisumbras, line 870: Rather than removing semicolon here and in other 100500 places, remove it in the macro definition, please.

nigels-com: Ah.  The macro itself doesn't seem to be the problem.  It's ending with a function, which ought not end with a semicolon.

```
#define BOOST_JSON_HANDLE_EVENT(fn) \
    struct do_ ## fn \
    { \
        template< class H, class... Args > \
        bool operator()( H& h, Args& ... args ) const \
        { \
            return h. fn (args...); \
        } \
    }; \
       \
    template< class... Args > \
    bool fn( error_code& ec, Args&& ... args ) \
    { \
        if( inner_active_ < 0 ) \
        { \
            BOOST_JSON_FAIL( ec, error::not_array ); \
            return false; \
        } \
        constexpr int N = std::tuple_size<T>::value; \
        if( inner_active_ >= N ) \
        { \
            BOOST_JSON_FAIL( ec, error::size_mismatch ); \
            return false; \
        } \
        using F = handler_op_invoker< do_ ## fn, error_code, Args...>; \
        using H = decltype(handlers_); \
        return mp11::mp_with_index<N>( \
            inner_active_, \
            tuple_handler_op_invoker<H, F>{ \
                handlers_, \
                F{ std::forward_as_tuple(ec, args...) } } ); \
    }

    BOOST_JSON_HANDLE_EVENT( on_object_begin );
    BOOST_JSON_HANDLE_EVENT( on_object_end );

```

grisumbras: Oh, sorry, I was confused.

grisumbras:
Merged via 942bbd4cf5e5fd30b7be71d187ab5bd2e54f8939
Thank you for your contribution!

Issue #940, opened 2023-09-20
 strings generated by the browser’s JSON.stringify are not parseable by Boost JSON


vaishnavkatiyar:
Some strings generated by the browser’s JSON.stringify are not parseable by Boost JSON in Boost 1.78.0

Browsers can generate JSON strings that look like (CPP string code):
 
    std::string jsonString = "{\"command\":\"[\\uD83D\\uD83D\](file://ud83d/uD83D/)",\"id\":\"id\"}";     // Illegal trailing surrogate
    std::string jsonString = "{\"command\":\"[\\uDF3E\\uDEC2\](file://udf3e/uDEC2/)",\"id\":\"id\"}";   // Illegal leading surrogate
    std::string jsonString = "{\"command\":\"[\\uD83D\](file://ud83d/)",\"id\":\"id\"}";   // Syntax Error (Half a Surrogate)

The above JSON strings are produced by the following JS when run in a modern browser:
JSON.stringify({ command: "\uD83D\uD83D", id: "id" });
JSON.stringify({ command: "\uDF3E\uDEC2", id: "id" });
JSON.stringify({ command: "\uD83D", id: "id" });

This JSON itself is valid UTF-8 strings, but when the strings in the JSON are parsed, and de-escaped the resulting strings are not valid UTF8, and Boost reports parsing errors.

The JSON is produced by browsers as part of the well-formed string specification: https://github.com/tc39/proposal-well-formed-stringify. 
 
The Boost Library provides an option called: boost_parse_options.allow_invalid_utf8 = true;
But this has no effect on the output for these strings. Boost still rejects this valid JSON.  We also checked with the latest version of Boost and it does not appear this is fixed.  

It would be helpful if Boost JSON was able to at least read the other fields, and report an issue for the field(s) that do not parse.

A  small program demonstrating the issue:
https://godbolt.org/z/eK6xa89x8

[boost_json_cpp.txt](https://github.com/boostorg/json/files/12677486/boost_json_cpp.txt)

grisumbras:
So, first of all "\uD83D\uD83D" is *not* a valid UTF-8 string. It does not represent any sequence of unicode characters. Whatever gives you strings like that does not give you UTF-8 encoded strings. My intuition is that you need to escape the slashes prior to `JSON.stringify`.

Second, `parse_options::allow_invalid_utf8` is intended to disable checking if UTF-8 code units in the text are forming a valid code point. That is, it affects how _unescaped_ UTF-8 is treated.

[...]

PR #985, opened 2024-02-28
Introduce allow_invalid_utf16 option in Boost JSON

vaishnavkatiyar:
Fix: [#940 ](https://github.com/boostorg/json/issues/940)

**Motivation**
Boost library provides parsing and serialization algorithms to transform JSON to and from the [value](https://www.boost.org/doc/libs/develop/libs/json/doc/html/json/ref/boost__json__value.html) container as needed.
The current version of the boost JSON library includes an option parse_options::allow_invalid_utf8 to control the treatment of unescaped UTF-8 code units, allowing users to disable the validation of UTF-8 code points.
It lacks a similar capability for dealing with invalid UTF-16 surrogate pairs. This limitation becomes apparent when users encounter scenarios involving UTF-16 encoded data with mismatched or out-of-range surrogate pairs, where the existing functionality does not provide adequate support or validation mechanisms.
Therefore, extending this functionality to encompass invalid UTF-16 surrogate pairs is necessary to ensure comprehensive handling of different UTF encoding scenarios and enhance the library's robustness and versatility in parsing and serialization tasks involving UTF-16 encoded JSON data.

**Design**
The introduction of the parse_options::allow_invalid_utf16 option grants the allowance of invalid UTF-16 surrogate pairs in strings processed by the parser.
When the option is activated, the parser refrains from raising errors for invalid UTF-16 sequences. Instead, it replaces these invalid sequences with the Unicode replacement character.

### DIFF
--- a/include/boost/json/basic_parser_impl.hpp
+++ b/include/boost/json/basic_parser_impl.hpp
@@ -563,39 +563,72 @@ do_doc1:
 do_doc2:
     switch(+opt_.allow_comments |
         (opt_.allow_trailing_commas << 1) |
-        (opt_.allow_invalid_utf8 << 2))
+        (opt_.allow_invalid_utf8 << 2) |
+        (opt_.allow_invalid_utf16 << 3))
     {
     // no extensions
     default:
-        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::false_type(), std::false_type());
+        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::false_type(), std::false_type(), std::false_type());
         break;
     // comments
     case 1:
-        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::false_type(), std::false_type());
+        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::false_type(), std::false_type(), std::false_type());
         break;
     // trailing
     case 2:
-        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::true_type(), std::false_type());
+        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::true_type(), std::false_type(), std::false_type());
         break;
     // comments & trailing
     case 3:
-        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::true_type(), std::false_type());
+        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::true_type(), std::false_type(), std::false_type());
         break;
     // skip validation
     case 4:
-        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::false_type(), std::true_type());
+        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::false_type(), std::true_type(), std::false_type());
         break;
     // comments & skip validation
     case 5:
-        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::false_type(), std::true_type());
+        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::false_type(), std::true_type(), std::false_type());
         break;
     // trailing & skip validation
     case 6:
-        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::true_type(), std::true_type());
+        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::true_type(), std::true_type(), std::false_type());
         break;
     // comments & trailing & skip validation
     case 7:
-        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::true_type(), std::true_type());
+        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::true_type(), std::true_type(), std::false_type());
+        break;
+    // allow_invalid_utf16
+    case 8:
+        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::false_type(), std::false_type(), std::true_type());
+        break;
+    // comments & allow_invalid_utf16
+    case 9:
+        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::false_type(), std::false_type(), std::true_type());
+        break;
+    // trailing & allow_invalid_utf16
+    case 10:
+        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::true_type(), std::false_type(), std::true_type());
+        break;
+    // skip validation & allow_invalid_utf16
+    case 11:
+        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::false_type(), std::true_type(), std::true_type());
+        break;
+    // comments & trailing & allow_invalid_utf16
+    case 12:
+        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::true_type(), std::false_type(), std::true_type());
+        break;
+    // comments & skip validation & allow_invalid_utf16
+    case 13:
+        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::false_type(), std::true_type(), std::true_type());
+        break;
+    // trailing & skip validation & allow_invalid_utf16
+    case 14:
+        cs = parse_value(cs.begin(), stack_empty, std::false_type(), std::true_type(), std::true_type(), std::true_type());
+        break;
+    // comments & trailing & skip validation & allow_invalid_utf16
+    case 15:
+        cs = parse_value(cs.begin(), stack_empty, std::true_type(), std::true_type(), std::true_type(), std::true_type());
         break;
     }
     if(BOOST_JSON_UNLIKELY(incomplete(cs)))
@@ -631,7 +664,8 @@ parse_value(const char* p,
     std::integral_constant<bool, StackEmpty_> stack_empty,
     std::integral_constant<bool, AllowComments_> allow_comments,
     /*std::integral_constant<bool, AllowTrailing_>*/ bool allow_trailing,
-    /*std::integral_constant<bool, AllowBadUTF8_>*/ bool allow_bad_utf8)
+    /*std::integral_constant<bool, AllowBadUTF8_>*/ bool allow_bad_utf8,
+    bool allow_bad_utf16)
 {
     if(stack_empty || st_.empty())
     {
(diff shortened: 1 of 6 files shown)


grisumbras:
It usually appears within an hour. I'm not sure performance will be affected, given that this is a special case within a special case.

include/boost/json/basic_parser_impl.hpp
```diff
@@ -1202,6 +1240,7 @@ parse_escaped(
     int digit;
     char c;
     cs.clip(temp.max_size());
+    bool replaced_bad_utf16 = false;
```
grisumbras, line 1243: Shouldn't this variable be a data member of the parser, so that suspensions doesn't lose its state?

include/boost/json/basic_parser_impl.hpp
```diff
@@ -563,39 +563,72 @@ parse_document(const char* p,
 do_doc2:
     switch(+opt_.allow_comments |
         (opt_.allow_trailing_commas << 1) |
-        (opt_.allow_invalid_utf8 << 2))
+        (opt_.allow_invalid_utf8 << 2) |
+        (opt_.allow_invalid_utf16 << 3))
```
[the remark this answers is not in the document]
vaishnavkatiyar: Passed it directly. Should I merge other cases by passing `opt_.allow_trailing_commas & opt_.allow_invalid_utf8` directly?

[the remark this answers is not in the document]
grisumbras: No, don't change old code.

include/boost/json/basic_parser_impl.hpp
vaishnavkatiyar: Added as a data member to the parser

include/boost/json/basic_parser_impl.hpp
```diff
@@ -1353,20 +1358,53 @@ parse_escaped(
...
+                {
+                    BOOST_STATIC_CONSTEXPR source_location loc
+                        = BOOST_CURRENT_LOCATION;
+                    return fail(cs.begin(), error::syntax, &loc);
+                }
+                // Otherwise, append the Unicode replacement character if
+                // the first code point is a valid leading surrogate.
+                else
+                {
+                    if(replaced_bad_utf16 == false)
```
grisumbras, line 1397: As previously mentioned, if a bad surrogate has been encountered, at this point you should have already been broken out of the switch. Instead, here you should be _replacing_ the contents of `temp` with UTF-8-encoded `0xFFFD`, because, if you are here, than the previous surrogate was good, but now you know it's not a part of a pair.

include/boost/json/basic_parser_impl.hpp
```diff
@@ -1401,16 +1439,46 @@ parse_escaped(
             unsigned const u2 =
                 (d1 << 12) + (d2 << 8) +
                 (d3 << 4) + d4;
-            // valid trailing surrogates are [DC00, DFFF]
-            if(BOOST_JSON_UNLIKELY(
-                u2 < 0xdc00 || u2 > 0xdfff))
+            // Append the second Unicode replacement character if
+            // the first code point is an illegal leading surrogate
+            if(replaced_bad_utf16 == true)
```
grisumbras, line 1444: Given that `break` I asked for is added, I don't think this would be necessary.

grisumbras (changes requested):
In addition to algorithmic changes I requested, please check that you don't create trailing whitespace anywhere.

include/boost/json/basic_parser_impl.hpp
```diff
@@ -1353,20 +1358,53 @@ parse_escaped(
...
+                    return fail(cs.begin(), error::illegal_leading_surrogate,
+                        &loc);
+                }
+                // Otherwise, append the Unicode replacement character and
+                // mark the surrogate as replaced.
+                else
+                {
+                    unsigned cp = 0xFFFD; // Unicode replacement character
+                    temp.append_utf8(cp);
+                    replaced_bad_utf16 = true;
```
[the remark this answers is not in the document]
vaishnavkatiyar: After making the suggested changes, the result looks like:
•	\uD800 results in U+FFFD (single high surrogate).
•	\uDC00 results in U+FFFD (single low surrogate).
•	\uDC00\uD800 results in U+FFFD U+FFFD (low surrogate followed by high surrogate).
•	\uDC00\uDC00 results in U+FFFD U+FFFD (low surrogate followed by low surrogate).
•	\uD800\uD800 results in U+FFFD U+FFFD (high surrogate followed by high surrogate).
•	\uD800\uDC00 is mapped to U+10000 (high surrogate followed by low surrogate, forming a valid surrogate pair).
•	\uDC00\uD800\uDC00 results in **U+FFFD U+10000** (evaluated as single low surrogate, high surrogate followed by low surrogate, forming a valid surrogate pair).
•	\uD800\uDC00\uDC00 results in **U+10000 U+FFFD** (evaluated as high surrogate followed by low surrogate, forming a valid surrogate pair, followed by a single low surrogate).

I have a confusion for the following use case:
•	\uD800\uD800\uDC00 results in **U+FFFD U+FFFD U+FFFD** (evaluated as high surrogate followed by high surrogate, forming an  invalid surrogate pair followed by single low surrogate).
It is not evaluated as **U+FFFD U+10000 ** since the first code point is a valid high surrogate and after checking for '**\**' and '**u**' the '**cs**' is pointing to the start of the second code point (**D**). Since we can't move the '**cs**' backwards to '**\**' breaking the switch here will not suffice for checking the second code point (**\uD800**) as a first surrogate for the subsequent code point (**\uDC00**).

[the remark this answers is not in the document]
grisumbras: Yeah, we can't move cs back, but we do have the value of the second code code point. This means we need to add code that checks if the code point was a valid high surrogate. And if it is, fill `temp` and go back to expecting the second surrogate.

include/boost/json/basic_parser_impl.hpp
```diff
@@ -1398,19 +1428,43 @@ parse_escaped(
...
+                if (u2 < 0xd800 || u2 > 0xdbff)
+                {
+                    temp.append_utf8(u2);
+                    break;
+                }
+                // If it is a valid leading surrogate
+                else
+                {
+                    u1 = u2;
+                    goto do_str9;
```
grisumbras, line 1462: If you go there you potentially violate the assumption that `cs.remain() > 10`, and so you can overrun the input buffer.

include/boost/json/basic_parser_impl.hpp
```diff
@@ -1576,13 +1654,38 @@ parse_escaped(
         }
         ++cs;
         u2_ += digit;
+        // Check if the second code point is a valid trailing surrogate.
+        // Valid trailing surrogates are [DC00, DFFF]
         if(BOOST_JSON_UNLIKELY(
-            u2_ < 0xdc00 || u2_ > 0xdfff))
+                u2_ < 0xdc00 || u2_ > 0xdfff))
```
grisumbras, line 1660: Why extra indent here?

grisumbras (changes requested):

include/boost/json/basic_parser_impl.hpp
```diff
@@ -1353,20 +1358,45 @@ parse_escaped(
...
+                {
+                    BOOST_STATIC_CONSTEXPR source_location loc
+                        = BOOST_CURRENT_LOCATION;
+                    return fail(cs.begin(), error::syntax, &loc);
+                }
+                // Otherwise, append the Unicode replacement character if
+                // the first code point is a valid leading surrogate
+                else
+                {
+                    temp.append_utf8(urc);
```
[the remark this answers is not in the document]
vaishnavkatiyar: I think we are storing the value of the leading surrogate in **u1** and appending it to **temp** in line **1356** only (when it is a valid utf-8 code point).
In other cases, we are not appending **u1** to **temp**, hence, directly appending the **replacement character** when:
1. First code point is NOT a valid high surrogate line **1374**
2. First code point is a valid high surrogate but second code point is not present line **1397**
3. First code point is a valid high surrogate but second code point is not a valid low surrogate line **1450**

include/boost/json/basic_parser_impl.hpp
vaishnavkatiyar: Since the value of the second code point **u2** is not stored in temp, how can we preserve it if we don't use goto and break instead?

grisumbras: You can assign to `u1_` and go to `do_sur1`.

include/boost/json/basic_parser_impl.hpp
```diff
@@ -1353,20 +1359,44 @@ parse_escaped(
...
+                    break;
```
grisumbras, line 1398: Add a test that covers this case . Something like `\uD800X` should work.

grisumbras (changes requested):

include/boost/json/basic_parser_impl.hpp
```diff
@@ -1353,20 +1359,44 @@ parse_escaped(
...
+                }
             }
             ++cs;
             if(BOOST_JSON_UNLIKELY(*cs != 'u'))
```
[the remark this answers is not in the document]
grisumbras: Same for [line 1604](https://github.com/boostorg/json/pull/985/files#diff-ac0f59cf63d51320da1fc37682921e31221528ac2e1c8412363fa3f4d2dead5cR1604)

grisumbras (changes requested):

grisumbras (approved):
Looks alright. Now please rebase on the current develop branch and squash.

grisumbras:
Thank you for your work, I'll merge this soon.

vaishnavkatiyar:
Thank you @grisumbras for your continued support and valuable feedback.

vaishnavkatiyar:
Hello @grisumbras 
I hope this message finds you well. I wanted to inquire about the anticipated timeframe for the merging of this pull request. Your attention to this matter is greatly appreciated.
Thank you

grisumbras:
I need to finish my current project (which will probably happen tomorrow). After that I'll merge this. The change will definitely be in the next Boost release which will happen in August.

vaishnavkatiyar:
Thank you for the update. Looking forward to the next Boost release.

PR #1027, opened 2024-07-19
Docs: list bench command line options

sdarwin:
It's not always convenient to "run the bench program with no arguments for a list of command line options" if you want to review the command line options. That requires building and compiling everything.   How about including options in the README page.

One of those options is `-f   Include file IO into consideration when testing parsers`.  Could you double-check and confirm that ordinarily (without the -f option) the benchmarks avoid including file IO in the calculations?  Does that include both ingesting files prior to running the benchmark, as well as writing the results afterwards?  Both of those?  Therefore, when diagnosing the benchmark results the file system read/write time can be ignored?  A (variably) slow disk system would not affect the benchmarks?

### DIFF
--- a/bench/README.md
+++ b/bench/README.md
@@ -5,6 +5,28 @@ fetch the third party repositories. Then run
 the bench program with no arguments for a
 list of command line options.
 
+```
+Usage: bench [options...] <file>...
+
+Options:  -t:[p][s]            Test parsing, serialization or both
+                                 (default both)
+          -i:[b][d][r][c][n]   Test the specified implementations
+                                 (b: Boost.JSON, pool storage)
+                                 (d: Boost.JSON, default storage)
+                                 (u: Boost.JSON, null parser)
+                                 (s: Boost.JSON, convenient functions)
+                                 (o: Boost.JSON, stream operators)
+                                 (default all)
+          -n:<number>          Number of trials (default 6)
+          -b:<branch>          Branch label for boost implementations
+          -m:(i|p|n)           Number parsing mode
+                                 (i: imprecise)
+                                 (p: precise)
+                                 (n: none)
+                                 (default imprecise)
+          -f                   Include file IO into consideration when testing parsers
+```
+
 When building with b2, it is possible to create several different copies of the
 bench program for different build properties (toolset, build variant, etc.).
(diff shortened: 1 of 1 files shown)


grisumbras:
`-f` uses a completely different function, so this mode should not affect the regular mode. In regular mode the benchmark reads the file, and then uses this information to parse/serialize the document without suspensions. The original read does not affect that calculation (in fact, the code written in a way, so it cannot affect it). With `-f` Boost.JSON implementations  use partial reads and partial parses. I introduced this mode in order to test two things:

1. How comparable is it to use verbose approach of setting up a streaming parser, as opposed to just using iostreams.
2. Whether it is better to read a file entirely into memory, then parse it without suspensions, or to parse available chunks.

When counting the time taken by the functions in `-f` mode the time spent on opening, reading, closing the file, as well as on parsing it, is counted. For each iteration. Priniting the results is not counted. I haven't implemented this mode for serialization, as it opens the question of where to write the data to. This can be affected by slow FS, and disk caching. So, admittedly, it's not a very representative benchmark.

Regular mode should not be affected by FS operations, as far as I can tell. If it is, it's definitely a bug.

sdarwin:
> Regular mode should not be affected by FS operations, as far as I can tell.

Yes ok.  Since the existing benchmarks are stable it seems probable that filesystem operations aren't interfering.

sdarwin:
The new benchmark runner is an Intel(R) Xeon(R) E-2224 CPU @ 3.40GHz from knownhost.com.  First released in May 2019.

The intel i5 from last week was consistently averaging 30-35 out of 56 benchmarks differing from a perfect measurement of 0% .

The results in this pull request on the E-2224 shows 17 out 56 benchmarks differing from 0%, which is significantly better than the i5.    However, still worse than the earlier Intel(R) Xeon(R) CPU E5620  @ 2.40GHz.     It seems that less is more.  As Intel designs more power management features in their newer CPUs, that comes at a cost of fluctuations and variability.

The E-2224 is now "in production" for boostorg/json, let's observe how the benchmarks go.

grisumbras:
So, can I merge this?

sdarwin:
Yes

grisumbras:
Merged via 56f3e1716345f5ceb2f03e78ba9f6b803990d034

PR #1038, opened 2024-09-01
extend fuzzing with different parse options

tyler92:
Include different parse_options for fuzzing for better coverage


grisumbras:
Good idea. Thank you for your contribution.

PR #1043, opened 2024-09-19
Allow to include pretty printers only once

Julien-Blanc-tgcm:
* some linkers will include the section many times in the executable, resulting in increased load time and warning messages inside gdb
* allows including them only once by defining BOOST_ALL_NO_EMBEDDED_GDB_SCRIPTS, and including the printers.hpp file only once.

### DIFF
--- a/include/boost/json/debug_printers.hpp
+++ b/include/boost/json/debug_printers.hpp
@@ -7,7 +7,7 @@
 // Official repository: https://github.com/boostorg/json
 //
 
-#ifndef BOOST_JSON_PRINTERS_HPP
+#ifndef BOOST_JSON_DEBUG_PRINTERS_HPP
 
 #include <boost/json/detail/gdb_printers.hpp>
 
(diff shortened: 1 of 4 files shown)


Julien-Blanc-tgcm:
I think this feature should be documented somewhere, which is missing currently. Which part of the doc would fit ?

grisumbras:
It's hard to say. I planned to write a dedicated section on error handling, debugging and so on. I'll describe your technique there.

grisumbras:
Is there any reason to have both `json/printers.hpp` and `json/detail/gdb_printers.hpp`?

Julien-Blanc-tgcm:
I don't like the idea of asking users to explicitly include a file in the detail directory. Also, i think it is better not to name the file gdb, as we could include more printers there (ie, for visual studio).

However i realize that calling it just printers.hpp may be misleading. Perhaps i should rename it to debug_printers.hpp .

grisumbras:
`debug_printers.hpp` sounds all right.

grisumbras:
Merged via 743be7fe13bbea76f651ab81703a5b8e1fdf25e0

Issue #1048, opened 2024-10-06
parse_into.hpp:109: Assertion `false' failed with std::tuple


tyler92:
### Version of Boost

`devel` branch, commit `85f9e92c18283d42737ebd067df4f881824e4edb`

### Steps necessary to reproduce the problem

```cpp
#include <boost/json/parse_into.hpp>
#include <boost/describe/class.hpp>

struct Test
{
  std::tuple<bool, int> t1;
};

BOOST_DESCRIBE_STRUCT(Test, (), (t1))

int main()
{
  Test object{};
  std::error_code ec;

  std::string data = R"({"t1":[]})";
  boost::json::parse_into(object, data, ec);

  return 0;
}
```

Output:

```
test-parse: boost/json/detail/parse_into.hpp:109: bool boost::json::detail::handler_error_base<boost::json::error::not_bool>::on_object_end(system::error_code &) [E = boost::json::error::not_bool]: Assertion `false' failed.
Aborted (core dumped)
```

### All relevant compiler information

Clang version 17.0.6

grisumbras:
Thank you for catching this!

Issue #1047, opened 2024-10-06
Buffer overflow in boost::json::parse_into with std::array


tyler92:
### Version of Boost

`devel` branch, commit `85f9e92c18283d42737ebd067df4f881824e4edb`

### Steps necessary to reproduce the problem

```cpp
#include <boost/json/parse_into.hpp>
#include <boost/describe/class.hpp>

struct Test
{
  std::array<std::int64_t, 3> arr;
};

BOOST_DESCRIBE_STRUCT(Test, (), (arr))

int main()
{
  Test object{};
  std::error_code ec;

  std::string data = R"({"arr":[977,775500052916,9216,77552916,9216]})";
  boost::json::parse_into(object, data, ec);

  return 0;
}
```

### All relevant compiler information

Address sanitizer report a buffer overflow here:

```
=================================================================
==233641==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7f5a9fc00038 at pc 0x559dfb6751d6 bp 0x7ffd547ca3a0 sp 0x7ffd547ca398
WRITE of size 8 at 0x7f5a9fc00038 thread T0
    #0 0x559dfb6751d5 in boost::json::detail::converting_handler<boost::json::detail::sequence_conversion_tag, std::array<long, 3ul>, boost::json::detail::converting_handler<boost::json::detail::described_class_conversion_tag, Test, boost::json::detail::into_handler<Test>>>::signal_value() boost/json/detail/parse_into.hpp:567:21
    #1 0x559dfb674f9f in boost::json::detail::converting_handler<boost::json::detail::integral_conversion_tag, long, boost::json::detail::converting_handler<boost::json::detail::sequence_conversion_tag, std::array<long, 3ul>, boost::json::detail::converting_handler<boost::json::detail::described_class_conversion_tag, Test, boost::json::detail::into_handler<Test>>>>::on_int64(boost::system::error_code&, long) boost/json/detail/parse_into.hpp:298:24
[...]

PR #1052, opened 2024-10-08
direct parsing fuzzing test

tyler92:
Two issues were found by the new fuzzing target: https://github.com/boostorg/json/issues/1047 and https://github.com/boostorg/json/issues/1048

I think detecting memory issues in case of new changes or refactoring might be useful.
I'm not sure about code formatting for the project, I will appreciate your feedback.


grisumbras:
While the idea is interesting, I am sceptical fuzzing with random strings can be effective when the parser expects a very specific structure of the input. It will probably always fail on the second (`on_object_begin`), the third (`on_key`), or the fourth (`on_bool`) event.

tyler92:
According to coverage reports, it works well enough to find valid inputs, these two issues were found quite quickly. We can provide a dictionary for the fuzzer to help it a bit if we want to improve the performance.

sdarwin:
It may be a coincidence. This pull request mentions "detecting memory issues". The drone agent on IBM s390 crashed with out-of-memory errors.  I have rebooted the server.

grisumbras:
It's also possible that linking the new fuzzer requires too much memory (variant parsers do do that).

tyler92:
> I am sceptical fuzzing with random strings can be effective when the parser expects a very specific structure of the input. It will probably always fail on the second (on_object_begin), the third (on_key), or the fourth (on_bool) event.

Just for an experiment: I reverted the fix for https://github.com/boostorg/json/issues/1047 and measured the time required to find that bug. Without a dictionary and with an empty corpus (the worst scenario) it takes about 10 seconds. The reason is the fuzzer doesn't have to find a valid input with all required fields. It's enough to generate a JSON with some fields to verify parser for a specific type only.

Adding corpus doesn't improve speed much. But I've added two valid JSONs to seed files, and it gives the fuzzer the ability to reach the "success" point. With this change, the fuzzer finds https://github.com/boostorg/json/issues/1047 in less than one second. Also, I changed C++ version to C++14 to avoid potential OOM.

Please let me know if it makes sense

fuzzing/fuzz_direct_parse.cpp
```diff
@@ -0,0 +1,92 @@
...
+    std::array<std::uint64_t, 3> a3;
+    std::map<std::string, std::int64_t> m1;
+    std::map<std::string, std::string> m2;
+    std::map<std::string, double> m3;
+    std::tuple<bool, std::uint64_t, std::int64_t, double, std::string> t1;
+    std::tuple<std::array<std::string, 3>, std::array<double, 3>, std::nullptr_t> t2;
+    std::tuple<std::vector<std::string>, std::vector<double>> t3;
+
+#ifndef BOOST_NO_CXX17_HDR_VARIANT
+    std::variant<bool, std::uint64_t, std::int64_t, double, std::string> v;
```
grisumbras, line 51: Please, replace with `boost::variant2::variant`. It will reduce the amount of macro switching necessary.

.github/workflows/run_fuzzer.yml
```diff
@@ -60,6 +60,7 @@ jobs:
         buildtype: 'boost'
         path: 'head'
         toolset: clang-18
+        cxxstd: 14
```
grisumbras, line 63: Why have you reduced this from 17 to 14?

fuzzing/fuzz_direct_parse.cpp
tyler92: Done

.github/workflows/run_fuzzer.yml
tyler92: I wanted to make sure this MR has a chance to be approved and avoid potential OOMs on the drone agent before (although I'm not sure if it's related to the current MR). I reverted to 17; please let me know if there are any issues with it.

grisumbras (approved):
Everything looks good. Can you please squsah and rebase on the current develop (if it's not already). I can also do it manually myself, but then the PR won't be considered merged by GitHub.

tyler92:
> Can you please squsah and rebase on the current develop (if it's not already). I can also do it manually myself, but then the PR won't be considered merged by GitHub.

Done

grisumbras:
Thank you for your contribution.

PR #1070, opened 2025-02-11
Boost Intrusive to be able to print the json pointer of where an error occourred when performing the value_to operation.

RoyBellingan:
When used with

https://github.com/RoyBellingan/boostIntrusive/tree/master

It should produce this

```


------
default msg : source composite size does not match target size
expanded msg: the key >> timestamp << is non optional and missing in path /0/comment/0

------
default msg : source composite size does not match target size
expanded msg: the key >> author << is non optional and missing in path /0/comment/0/likes/1

------
default msg : not a number
expanded msg: Last json pointer was /0/comment/0/timestamp

```


grisumbras:
I have an idea for an alternative approach. I'll implement it first, so that I can compare yours with mine.

grisumbras:
My alternative: #1075.

PR #1073, opened 2025-02-15
missing BOOST_FALLTHROUGH in a few cases

RoyBellingan:
After seeing a few test failing for #1071 I added a few more check in my local build, and those 3 popped out

### DIFF
--- a/include/boost/json/basic_parser_impl.hpp
+++ b/include/boost/json/basic_parser_impl.hpp
@@ -686,7 +686,7 @@ loop:
             // or just leave it as is
             if(BOOST_JSON_UNLIKELY(p == sentinel()))
                 return maybe_suspend(p, state::val2);
-            // intentional fallthrough
+            BOOST_FALLTHROUGH;
         case ' ':
         case '\t':
         case '\n':


include/boost/json/basic_parser_impl.hpp
```diff
@@ -686,10 +686,17 @@ parse_value(const char* p,
             // or just leave it as is
             if(BOOST_JSON_UNLIKELY(p == sentinel()))
                 return maybe_suspend(p, state::val2);
-            // intentional fallthrough
+            BOOST_FALLTHROUGH;
+
         case ' ':
+            BOOST_FALLTHROUGH;
```
grisumbras, line 692: This and the next few are not necessary. It's a common idiom to group cases like this. I don't believe any compiler would complain.

grisumbras (changes requested):

include/boost/json/basic_parser_impl.hpp
grisumbras: GCC doesn't seem to issue a warning there even with the explicit flag. To be honest, marking every empty case like this looks very ugly, so I'm going to bite the bullet and insist you remove those lines.

RoyBellingan:
Each Journey begin with one step!

PR #1075, opened 2025-02-25
basic implementation of context-based path traction

grisumbras:
Very rough draft of the idea.

For a context type Ctx the user is supposed to implement these functions:
```
void tag_invoke(context_key_push_tag, string_view,  Ctx const&);
void tag_invoke(context_key_pop_tag, Ctx const&);
void tag_invoke(context_index_push_tag, std::size_t, Ctx const&);
void tag_invoke(context_index_pop_tag, Ctx const&);
```

The library would invoke them when traversing through the converted/parsed/serialized value.

Things to think about:

* The user may actually want to set `error_code` in those callbacks.
* The user may want to clear `error_code` in those functions. But then where would we get the results the calling function was supposed to create?

### DIFF
--- a/include/boost/json/conversion.hpp
+++ b/include/boost/json/conversion.hpp
@@ -12,6 +12,7 @@
 
 #include <boost/json/detail/config.hpp>
 #include <boost/json/fwd.hpp>
+#include <boost/json/string_view.hpp>
 
 #include <type_traits>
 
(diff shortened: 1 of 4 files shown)


RoyBellingan:
Is very interesting and clean.
I would suggest in this section

https://github.com/grisumbras/json/blob/e4e192a84f7482aaad923575e34e6f37669252d0/include/boost/json/detail/value_to.hpp#L376

To move the 
  json::context_key_push(ctx, D::name);
  
Before the check, so we can capture the "last processed" json pointer

PR #1078, opened 2025-02-27
CI improvements

grisumbras:


grisumbras:
@sdarwin can you switch s390 machines back to the normal amount of jobs? I changed CI script to launch the heavy job separately, which should avoid memory exhaustion.

sdarwin:
> launch the heavy job separately

has it been merged?  
will that affect the 18 open pull requests still containing the earlier ci files? maybe not.
It's safer to keep the settings for now.  At least jobs will run faster since they have the full machine CPU available.

grisumbras:
Sure, but I can't check if this actually solves the problem, unless CI run it.

sdarwin:
It's possible to observe the memory usage, this PR does help. Notice it doesn't peak at 16GB and crash.

Screenshot from 2025-02-27 08-11-25

grisumbras:
Lol, I see, thanks.

PR #1080, opened 2025-04-11
New documentation formatting

grisumbras:


grisumbras:
@sdarwin I'm switching JSON docs to asciidoctor. I've changed the image that tests docs to the same one used for Boost releases, and also changed the script. Can you update the preview script too? Or ideally, we can come up with a way to control how the preview is generated from the project sources themselves.

sdarwin:
Hi @grisumbras , no changes in doc generation are needed.  :-)     

Some details to consider though.  

Previous preview link:   
https://1080.json.prtest2.cppalliance.org/libs/json/doc/html/index.html
New preview link:   
https://1080.json.prtest2.cppalliance.org/libs/json/doc/html/json/index.html

Since past versions of boost were using doc/html/index.html , if you really plan to modify the path of the docs,  place a static html file (or a dynamically generated one) at the `doc/html/index.html` based on this example:  https://github.com/boostorg/json/blob/develop/index.html  If anyone fiddles with the URL in the browser and visits the previous `doc/html/index.html` they will get redirected to the new doc location.

Even better would be to consider the example of boostorg/charconv. Here is a preview URL:

https://228.charconv.prtest.cppalliance.org/libs/charconv/doc/html/index.html

In other words, charconv doesn't add an extra layer of redirection ( an extra /charconv/  in the path ) and continues to use `doc/html/index.html` . Could that be done?

PR #1091, opened 2025-07-02
raise priority of optional and variant above sequence

### LINKED ISSUE #1050: Handling optional-like sequences
As Andrzej pointed out recently in a draft paper on the Library Evolution reflector, the JSON conversion heuristics currently prioritize checking sequence-like before checking optional-like. This means that any optional-like types that are ranges will serialize like a sequence (`[42]` and `[]` instead of `42` and `null`) and fail to deserialize (due to lack of `push_back` or `insert`), and, more importantly, an optional type that later adopts a range interface will break any users that use Boost.JSON (without even knowing if any such users exist). 

There are [multiple](https://www.scala-lang.org/api/2.13.3/scala/Option.html) [languages](https://doc.rust-lang.org/std/option/enum.Option.html) whose optional type is a sequence, so it's not far-fetched to suggest that a C++ one might exist also (regardless of whether `std::optional` ends up being one or not).

There are at least two ways to preemptively address this issue that I can think of (though there are probably more solutions):

1. Prioritize checking optional-like over sequence-like. 
2. Have sequence-like just assert against optional-like (on the premise that you meet two criteria we don't know what you mean, so we're not going to guess - this would make (de)serialization ill-formed, but ill-formed is better than well-formed and wrong)

I think an `optional<pair<string, T>>` that is a range is also [very nearly map-like](https://godbolt.org/z/MWors36Pc), only rejected by `emplace` probably not returning tuple-like. But it hypothetically could also (maybe you return the previous state as a `bool`?), so some care needs to be taken there too. 

---

I think a similar question might be asked of described classes that are also ranges. How should those be serialized, as objects or as arrays?

grisumbras:
Fix #1050.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -8,6 +8,7 @@ include::doc/pages/definitions.adoc[]
 == Boost 1.89.0
 
 .API Changes
+* {issue}1050[#1050] Optionals that are ranges are categorised as optionals.
 * {issue}1086[#1086] Described class support is enabled for types with bases.
 
 == Boost 1.87.0
--- a/doc/pages/conversion/overview.adoc
+++ b/doc/pages/conversion/overview.adoc
@@ -70,6 +70,20 @@ a| The result is a number equal to input and has the type
 |The result is a <<ref_string>>.
 |The result is constructed from a <<ref_string_view>>.
 
+|Type satisfying <<ref_is_variant_like>>.
+|`std::variant` and similar types, e.g. `boost::variant2::variant`.
+|The result is equal to the result of conversion of the active variant
+    alternative.
+|The result holds the first alternative for which a conversion succeeds.
+
+|Type satisfying <<ref_is_optional_like>>
+|
+|If the input value is empty, the result is a `null`. Otherwise it is
+    equivalent to conversion of the object stored inside of optional.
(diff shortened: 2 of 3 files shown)


brevzin:
I don't think this is the right solution. It does directly solve this narrow problem, but it still does the wrong thing for [this example](https://github.com/boostorg/json/issues/1050#issuecomment-2397471194) 

I would still argue for specializing a [conversion kind](https://github.com/boostorg/json/issues/1050#issuecomment-2402422542) and having the automatic detection of kind fail if the type matches more than one category (because are you _sure_ you know which one the user meant?)

grisumbras:
> I would still argue for specializing a https://github.com/boostorg/json/issues/1050#issuecomment-2402422542 and having the automatic detection of kind fail if the type matches more than one category (because are you sure you know which one the user meant?)

I have another change that will add the ability to directly specialize the category (and deprecate the current category traits). But I don't think that default categorization should fail when several categories are matched. That would turn off default categorisation for `std::optional` and `std::array`. I think, it's okay to guess incorrectly if it's easily fixed by the user with one template specialisation. Looking back at filed issues I had, nobody (well, except for you in #1050)  has ever complained about a wrong categorization, but people have complained about things not getting categorized when they could have been.

PR #1100, opened 2025-08-21
Remove sprintf uses

grisumbras:


sdarwin:
@grisumbras 

You are solving the issue by "removing sprintf".   A google search shows sprintf is deprecated on macos. See attached file for a list of boost libraries that include it. Would you agree, or not: I should open github issues on those boost libraries. Is it reasonable to have "sprintf" in the code?

[sprintf.txt](https://github.com/user-attachments/files/21917730/sprintf.txt)

PR #1106, opened 2025-09-06
CMake boost super-build fixups for cmake-3.8

nigels-com:
I've been on a little bit of an adventure to see what version of cmake is required for building boost.
See [boostorg/boost#1081](https://github.com/boostorg/boost/pull/1081)

Json needs cmake-3.15.0, otherwise it errors with:

```
-- Configuring done
CMake Error at libs/json/test/CMakeLists.txt:63 (target_compile_options):
  Error evaluating generator expression:

    $<CXX_COMPILER_ID:GNU,Clang,AppleClang>

  $<CXX_COMPILER_ID> expression requires one or zero parameters.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@
 # Official repository: https://github.com/boostorg/json
 #
 
-cmake_minimum_required(VERSION 3.14...3.16)
+cmake_minimum_required(VERSION 3.8...3.16)
 
 set(BOOST_JSON_VERSION 2)
 if(BOOST_SUPERPROJECT_VERSION)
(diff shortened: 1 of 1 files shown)


Flamefire:
I can confirm this change is sound: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:CXX_COMPILER_ID
> Changed in version 3.15: Multiple compiler_ids can be specified. CMake 3.14 and earlier only accepted a single compiler ID.

nigels-com:
I've updated the change here to support cmake-3.14 (instead of cmake-3.15).

However cmake-3.13 cmake fails with `-DBUILD_TESTING=Y` due to parser requiring cmake-3.14:

```
$ PATH=/opt/cmake-3.13.0-Linux-x86_64/bin:$PATH cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DBUILD_TESTING=Y -DCMAKE_CXX_COMPILER=clang++-18 -DBUILD_TESTING=Y -DBOOST_INCLUDE_LIBRARIES=json && ninja -v
-- Boost: using system layout: include, bin, lib, lib/cmake
-- Boost: Release build, static libraries, MPI OFF, Python OFF, testing Y
-- Boost: libraries included: json
-- Boost.Charconv: quadmath support OFF
-- Boost.Context: architecture x86_64, binary format elf, ABI sysv, assembler gas, suffix .S, implementation fcontext
-- Boost.Fiber: NUMA target OS is linux
-- Boost.Iostreams: ZLIB ON, BZip2 ON, LZMA ON (multithreaded), Zstd ON
-- Found the following ICU libraries:
--   data (required)
--   i18n (required)
--   uc (required)
-- Found ICU: /usr/include (found suitable version "74.2", minimum required is "4.8.1") 
-- Boost.Locale: iconv ON, ICU ON, POSIX ON, std ON, winapi OFF
-- Boost.Math: standalone mode OFF
-- Boost.Multiprecision: standalone mode OFF
CMake Error at libs/parser/CMakeLists.txt:9 (cmake_minimum_required):
  CMake 3.14 or higher is required.  You are running version 3.13.0


-- Configuring incomplete, errors occurred!
```

Flamefire:
> The superproject cmake does seem to pulling in Parser via a GLOB in [BoostRoot.cmake](https://github.com/boostorg/cmake/blob/develop/include/BoostRoot.cmake#L244).
> 
> cmake-3.13 does succeed with `-DBOOST_EXCLUDE_LIBRARIES=parser` specified to skip parser specifically.

That glob is only for detecting available libraries. Libraries are only configured as requested. One exception is if `BUILD_TESTING` is set which adds all libraries but doesn't build them by default:
https://github.com/boostorg/cmake/blob/81b08178c629b94b67e3baff5b1b9bee25db7f3a/include/BoostRoot.cmake#L355-L378

PR #1129, opened 2025-12-04
Fix handling of unexpected arrays in nested described struct fields

freitass:
Fixes a corner case where, when parsing unexpected arrays in nested JSON objects, the `ignoring_handler` would incorrectly signal `signal_end` instead of `signal_value`, causing the parser to skip the rest of the containing object.

```cpp
struct X
{
    int a;
    float b;
    std::string c;
};

BOOST_DESCRIBE_STRUCT(X, (), (a, b, c))

struct W
{
    X x;
};

BOOST_DESCRIBE_STRUCT(W, (), (x))

// ...

const auto json = R"(
{
    "x": {
        "unexpected": [],
        "a": 1,
        "b": 3.14,
        "c": "hello"
    }
})";

W w{};
std::error_code ec;
parse_into(w, json, ec);

BOOST_TEST( !ec );               // Ok.
BOOST_TEST( w.x.a == 1 );        // Error: w.x.a is 0
BOOST_TEST( w.x.b == 3.14f );    // Error: w.x.b is 0.0
BOOST_TEST( w.x.c == "hello" );  // Error: w.x.c is ""
```

### DIFF
--- a/include/boost/json/detail/parse_into.hpp
+++ b/include/boost/json/detail/parse_into.hpp
@@ -1052,7 +1052,7 @@ struct ignoring_handler
         --array_depth_;
 
         if( (array_depth_ + object_depth_) == 0 )
-            return parent_->signal_end(ec);
+            return parent_->signal_value(ec);
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


grisumbras:
First, thank you for catching and fixing this.

Please, rebase on develop so that CI jobs succeed.

PR #1138, opened 2026-01-06
erase_at_pointer 

RoyBellingan:
Hopefully a more clear and decent version of  #1071

### DIFF
--- a/include/boost/json/impl/pointer.ipp
+++ b/include/boost/json/impl/pointer.ipp
@@ -520,6 +520,110 @@ value::set_at_pointer(
     return try_set_at_pointer(sv, ref, opts).value();
 }
 
+bool
+value::erase_at_pointer(
+    string_view         sv,
+    system::error_code& ec)
+{
+    ec.clear();
+    if(sv.empty()){
+        BOOST_JSON_FAIL(ec, error::syntax);
+        return false;
+    }
+
+    string_view previous_segment;
+
+    string_view segment = detail::next_segment(sv, ec);
+
+    auto result = this;
+    auto previous_result = this;
+
+    while (true)
+    {
+        if (ec.failed())
+            return false;
+
+        if (!result)
+        {
+            BOOST_JSON_FAIL(ec, error::not_found);
+            return false;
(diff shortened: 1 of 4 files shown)


include/boost/json/impl/pointer.ipp
```diff
@@ -520,6 +520,110 @@ value::set_at_pointer(
...
 }
 
+bool
+value::erase_at_pointer(
+    string_view         sv,
+    system::error_code& ec)
+{
+    ec.clear();
+    if(sv.empty()){
+        BOOST_JSON_FAIL(ec, error::syntax);
```
grisumbras: `error::syntax` is for JSON syntax errors. Arguably, `error::missing_slash` is fitting (you require at least one pointer token, and a pointer token starts with a /. But there's an interesting question if we want to allow zero tokens, and what to do in that case. For example, we can set the root object to null.

include/boost/json/impl/pointer.ipp
```diff
@@ -520,6 +520,110 @@ value::set_at_pointer(
...
+                auto const index = detail::parse_number_token(segment, ec);
+                segment          = detail::next_segment(sv, ec);
+                if (ec.failed())
+                    return false;
+                
+                auto& arr = result->get_array();
+                result    = arr.if_contains(index);
+                if( !result )
+                {
+                    BOOST_JSON_FAIL(ec, error::past_the_end);
```
grisumbras: `error::past_the_end` is for when the current token is the past-the-end token (`-`). This is just `error::not_found`.

grisumbras (changes requested):
In addition to other changes, I expect the tests to have the full code coverage of the added lines.

test/erase_at_pointer.cpp
```diff
@@ -0,0 +1,307 @@
...
+                ]
+            }
+        ]
+    }
+]
+
+)";
+        return boost::json::parse(raw);
+    }
+
```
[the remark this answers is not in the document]
RoyBellingan: 😞

PR #1151, opened 2026-03-23
Drone: VS2026

sdarwin:
If possible, fix before merging. The cmake jobs are not succeeding.


grisumbras:
Merged via 84d05110234bb64a44de40e8c17aed99ea6b87c2.

PR #1153, opened 2026-03-31
CI: platforms update

grisumbras:


sdarwin:
relocating this comment from slack to the PR:

I think this update is not following the intended idea.     Compare to boostorg/url.  The strategy is to install a very small matrix in .drone.star which calls generate().    However you don't need to define the entirety of generate() in drone.star.   If that should be updated, then submit a PR to cppalliance/ci-automation.    The generate() function lives in ci-automation and is shared among other repos.   You can still include custom jobs, beyond those.

However, just speculating, maybe that is what you are doing here already.  

- Debug the generate() function, if there is a problem
- send a PR to ci-automation. and then switch back to using that version.

sdarwin:
Ok, I hadn't examined it carefully. Perhaps you are completely re-writing generate() .  It's a totally different function.  Not a small bug fix. So the above comments are not valid.

grisumbras:
My goal is to write a generate function that I satsifies my needs then see if I can get that merged with the one in `cppalliance/ci-automation`. It has certain differences that I think makes it better than the original.

sdarwin:
Yet another implementation option: Flamefire's `job_impl()` function https://github.com/boostorg/boost-ci/blob/master/ci/drone/functions.star  

And pdimov's `linux_pipeline()` https://github.com/boostorg/unordered/blob/develop/.drone.jsonnet

sdarwin:
macos jobs aren't running.

Analysis:

Consider these functions: "linux_cxx","windows_cxx","osx_cxx","freebsd_cxx"
I am not sure how your code interacts with those.  If there is a strict wall between the new logic and these earlier functions, and it calls "osx_cxx" as before, you would expect it to work correctly. 

However it isn't.

If some variables are getting passed through, it could be an explanation for a problem.

Inside of osx_cxx:

```
    "platform": {
      "os": "darwin",
      "arch": arm64
    },
```

whereas you are defining:

```
            'platforms': {
                ('macos', 'x86_64'): {
                    'xcode_version': '15.4.0',
                },
```

The platform should be 'darwin' and not 'macos'.
The architecture should be  'arm64' and not 'x86_64'.

To repeat, if the "osx_cxx" is called as before, without interference I'd expect it to be fine, but if somehow the values 'macos' or 'x86_64' are bleeding through, that could break it.    Try switching the code to use 'darwin' 'arm64'.   

If you put back the previous matrix job osx_cxx("..") as-is, that would work, so the problem is somehow involved with the new logic.

PR #1159, opened 2026-05-05
Enable adding Ubuntu toolchain PPA by default

Flamefire:
The default in Boost.CI will be false but some jobs still rely on this, so keep it for now the default.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -7,7 +7,7 @@
 
 
 def main(ctx):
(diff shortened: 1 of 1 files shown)


grisumbras:
I decided to go the opposite way in #1158 and just not rely on the PPA by default.

Flamefire:
That's the best solution, yes.

PR #1161, opened 2026-05-10
Erase_at_pointer and related test

RoyBellingan:
based on the suggestion in https://github.com/boostorg/json/pull/1138

Splits the JSON Pointer via next_segment, then navigate to the parent with find_pointer, this avoid duplicating the walk_pointer logic.

### DIFF
--- a/include/boost/json/impl/pointer.ipp
+++ b/include/boost/json/impl/pointer.ipp
@@ -520,6 +520,84 @@ value::set_at_pointer(
     return try_set_at_pointer(sv, ref, opts).value();
 }
 
+bool
+value::erase_at_pointer (
+    string_view sv,
+    system::error_code& ec) noexcept
+{
+    ec.clear();
+    if(sv.empty()){
+        BOOST_JSON_FAIL(ec, error::missing_slash);
+        return false;
+    }
+
+    string_view walk = sv;
+    string_view last_segment;
+    while (true)
+    {
+        last_segment = detail::next_segment(walk, ec);
+        if (ec.failed())
+            return false;
+        if (walk.empty())
+            break;
+    }
+
+    string_view const parent_sv(
+        sv.data(),
+        static_cast<std::size_t>(last_segment.data() - sv.data()));
+
+    value* parent = detail::walk_pointer(
+        *this,
+        parent_sv,
+        ec,
+        []( object& obj, detail::pointer_token token )
+        {
+            return detail::if_contains_token(obj, token);
+        },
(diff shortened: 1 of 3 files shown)


include/boost/json/impl/pointer.ipp
```diff
@@ -520,6 +520,84 @@ value::set_at_pointer(
...
+            break;
+    }
+
+    string_view const parent_sv(
+        sv.data(),
+        static_cast<std::size_t>(last_segment.data() - sv.data()));
+
+    value* parent = detail::walk_pointer(
+        *this,
+        parent_sv,
```
grisumbras, line 551: Wouldn't it make more sense to just walk the entire pointer string and in object and array handlers update a `value* parent` variable (in the case of object you'd have to also store the key)? Then, after you finished walking:

* `parent` is a pointer to `array`, `object` or `nullptr` (you can rely on that later in the switch case).
* You don't have to check for missing slash manually.
* You don't have to walk the pointer string twice.
* You don't have to do manually look up the pointed-to element in the parent.

include/boost/json/impl/pointer.ipp
```diff
@@ -520,6 +520,84 @@ value::set_at_pointer(
...
+        ec,
+        []( object& obj, detail::pointer_token token )
+        {
+            return detail::if_contains_token(obj, token);
+        },
+        []( array& arr, std::size_t index, system::error_code& ec ) -> value*
+        {
+            if( ec )
+                return nullptr;
```
grisumbras, line 560: This is not covered by tests. See here: https://1161.json.prtest2.cppalliance.org/diff-report/include/boost/json/impl/pointer.ipp.html. There are also other lines that miss coverage.

include/boost/json/impl/pointer.ipp
```diff
@@ -520,6 +520,84 @@ value::set_at_pointer(
...
+        case boost::json::kind::array: {
+            auto const index = detail::parse_number_token(last_segment, ec);
+            auto& arr = parent->get_array();
+            if (arr.if_contains(index)){
+                arr.erase(arr.begin() + index);
+                return true;
+            }
+            return false;
+        }
+        default: {
```
grisumbras, line 593: https://1161.json.prtest2.cppalliance.org/diff-report/include/boost/json/impl/pointer.ipp.html#NL593

I am pretty sure this switch case is unreachable, so replace it with
```
default: // LCOV_EXCL_LINE
   BOOST_JSON_UNREACHABLE(); // LCOV_EXCL_LINE
```

grisumbras (changes requested):

PR #1166, opened 2026-07-06
clamp count in string_impl::replace_unchecked

Ramya-9353:
Repro: `boost::json::string("hello").replace(0, 100, 0, 'x')` — the count exceeds `size() - pos`.
Cause: `string_impl::replace_unchecked` doesn't clamp `n1` to `size() - pos`, so `curr_size - pos - n1 + 1` underflows and the `memmove` runs a near-`SIZE_MAX` length out of bounds (ASAN: `negative-size-param` at string_impl.ipp:409). Overload (4) `string::replace(pos, count, count2, ch)` forwards `count` unchanged, though it is documented to replace `std::min(count, size() - pos)` characters. The `string_view` overload already clamps via `string_impl::replace`.
Fix: clamp `n1` to `size() - pos`, matching `string_impl::replace`.

### DIFF
--- a/include/boost/json/detail/impl/string_impl.ipp
+++ b/include/boost/json/detail/impl/string_impl.ipp
@@ -395,6 +395,7 @@ replace_unchecked(
         detail::throw_system_error( error::out_of_range, &loc );
     }
     const auto curr_data = data();
+    n1 = (std::min)(n1, curr_size - pos);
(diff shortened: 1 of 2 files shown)


grisumbras:
Thank you for catching this.

Ramya-9353:
Thanks for the quick merge, glad it was straightforward to land.

PR #1167, opened 2026-07-07
copy fill value before reallocation in array::insert/resize

Ramya-9353:
Repro: on a full array, `a.insert(a.begin(), n, a[0])` or `a.resize(n, a[1])` reads freed memory (ASAN heap-use-after-free in the value copy constructor).
Cause: revert_insert relocates the elements into a newly allocated table and frees the old one before the fill loop runs, so a value that aliases an existing element is left dangling.
Fix: copy the value into a local before constructing revert_insert, the same snapshot object::emplace_impl and array::emplace already take. std::vector defines these self-referential fills, so the copy overloads should honour it too.

### DIFF
--- a/include/boost/json/impl/array.ipp
+++ b/include/boost/json/impl/array.ipp
@@ -494,11 +494,15 @@ insert(
     value const& v) ->
         iterator
 {
+    // v may refer to an element of this array, whose
+    // storage revert_insert can relocate and free, so
+    // copy it before inserting
+    value const tmp(v, sp_);
     revert_insert r(
         pos, count, *this);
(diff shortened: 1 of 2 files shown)


grisumbras:
Thank you for catching this. I suspect the ideal solution is actually to change the API to accept by value and then eliminate the problem entirely. But since we're hours before the beta cutoff, I'm going to merge this now.

Ramya-9353:
Thanks for merging. Agreed that taking the fill value by value would remove the aliasing hazard at the API level; the local copy was just the minimal change for now.

PR #1168, opened 2026-07-08
fix self-append use-after-free in string::append

Ramya-9353:
Repro: on a heap string with no spare capacity, `s += s` (equivalently `s.append(s)` or `s.append(s.subview(k))`).
Cause: append(string_view) copies from sv after string_impl::append has reallocated and freed the buffer sv points into, so the copy reads freed memory (ASAN heap-use-after-free). insert and replace already guard self-aliasing via detail::ptr_in_range; append was the gap.
Fix: detect the alias before appending and copy from the relocated offset in the resulting buffer.

Regression tests in test/string.cpp cover whole and partial self-append forcing reallocation; they report a use-after-free before the change and pass after.

### DIFF
--- a/include/boost/json/detail/string_impl.hpp
+++ b/include/boost/json/detail/string_impl.hpp
@@ -27,6 +27,12 @@ class string;
 
 namespace detail {
 
+bool
+ptr_in_range(
+    const char* first,
+    const char* last,
+    const char* ptr) noexcept;
+
 class string_impl
 {
     struct table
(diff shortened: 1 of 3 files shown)


grisumbras:
Thank you for catching this.

Merged as cdc33018b1cd67f24c34aa0ab495584bd42b9f82.

Ramya-9353:
Thanks for taking this one in, glad it landed.

PR #1170, opened 2026-07-10
buffer source before reallocation in string::append(InputIt, InputIt)

Ramya-9353:
Repro: on a heap string with no spare capacity, self-append through the iterator overload, e.g. `s.append(s.begin(), s.end())` or `s.append(s.subview(k).begin(), s.subview(k).end())`.
Cause: append(InputIt, InputIt) sends random-access iterators to a path that grows the buffer with string_impl::append and then std::copy(first, last, out). The append reallocates and frees the old buffer that [first, last) points into, so the copy reads freed memory (ASAN heap-use-after-free at string.hpp:206). The string_view and input-iterator overloads already avoid this; the random-access one copied straight from the aliased source.
Fix: when the append would reallocate, buffer the characters through the input-iterator path first, and keep the in-place copy when there is spare capacity. std::string defines this overload as append(basic_string(first, last)), so self-referential ranges are well defined.

Regression tests in test/string.cpp self-append through the iterator overload with capacity consumed so the append reallocates; they report a use-after-free before the change and pass after.

### DIFF
--- a/include/boost/json/impl/string.hpp
+++ b/include/boost/json/impl/string.hpp
@@ -195,24 +195,42 @@ append(
     InputIt last,
     std::random_access_iterator_tag)
 {
-
     auto const n = static_cast<
         size_type>(last - first);
-    if(n > impl_.capacity() - impl_.size())
+    auto const size = impl_.size();
+    detail::string_impl tmp;
+    char* p = impl_.data();
+    if(n > impl_.capacity() - size)
     {
-        // appending forces a reallocation that frees our storage; if
-        // [first, last) aliases it the copy below would read freed memory,
-        // so buffer the characters first
-        append(first, last, std::input_iterator_tag{});
-        return;
+        // [first, last) may alias the current
+        // storage, so copy the new characters
+        // into the new storage before freeing it
+        tmp = detail::string_impl(
+            detail::string_impl::growth(
+                size + n, impl_.capacity()),
+            sp_);
+        p = tmp.data();
     }
-    char* out = impl_.append(n, sp_);
+
+    char* out = p + size;
 #if defined(_MSC_VER) && _MSC_VER <= 1900
     while( first != last )
         *out++ = *first++;
 #else
     std::copy(first, last, out);
 #endif
+
+    if(p != impl_.data())
+    {
+        std::memcpy(p, impl_.data(), size);
+        tmp.term(size + n);
+        impl_.destroy(sp_);
+        impl_ = tmp;
+    }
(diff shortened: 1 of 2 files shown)


grisumbras:
Thank you for catching this, but I am not sold on this implementation. Degrading to InputIterator implementation is quite a pessimization.

What really should happen is the library should
1. Allocate new storage.
2. Copy the new characters into the corrsponding positions.
3. Copy old characters.
4. Swap storages and destroy the old storage.

Something like
```
auto const n = static_cast<size_type>(last - first);
char const* p = data();
string_impl tmp;
if( n > capacity() - size() )
{
   tmp = string_impl(size() + n, sp_);
   p = temp.data();
}

std::copy(firt, last, p + n);

if( p != data() )
{
    std::memcpy( p, data(), size() );
    impl_.destroy(sp_);
    impl_ = tmp;
}
```

Also, I feel like of the two new tests the first one is unnecessary.

Ramya-9353:
Reworked as suggested: on reallocation the new characters are copied into the new storage while the old buffer is still live, then the old characters, then the storages are swapped and the old one destroyed. One deviation from the sketch: the new capacity comes from string_impl::growth(size + n, capacity()) rather than size() + n, so the growth factor stays consistent with string_impl::append and repeated appends remain amortised. Also dropped the first test.

The string suite passes ASAN-clean, as does an aliasing stress (whole and partial self-append, with and without spare capacity, checked against std::string semantics).

include/boost/json/impl/string.hpp
```diff
@@ -195,16 +195,42 @@ append(
...
+    if(p != impl_.data())
+    {
+        std::memcpy(p, impl_.data(), size);
+        tmp.term(size + n);
+        impl_.destroy(sp_);
+        impl_ = tmp;
+    }
+    else
+    {
+        impl_.term(size + n);
```
grisumbras: Isn't this just the same as your line 226? How about 

```
if( p != impl_.data() )
{
    std::memcpy(p, impl_.data(), size);
    impl_.destroy(sp_);
    impl_ = tmp;
}
impl_.term(size + n);
```

Ramya-9353: Yes, once impl_ is swapped the two calls are identical, so a single term after the branch does it. Restructured as suggested; the string suite and the aliasing stress still pass ASAN-clean.

grisumbras:
Merged as 6a5777635f5e66bca39f384301d7d9dc014ee2a0.

PR #1171, opened 2026-07-13
fix overflow check in parse_number_token

Ramya-9353:
value{{"foo", {1, 2, 3}}}.find_pointer(
        "/foo/" + std::to_string(SIZE_MAX) + "9", ec);
    // ec == error::not_found, expected error::token_overflow

parse_number_token flags `std::size_t` overflow with `new_result < result`, the
addition-overflow idiom, but the accumulation is `result * 10 + d`. Once
`result * 10` wraps, the wrapped value plus `d` can land at or above `result`,
so a pointer token an order of magnitude past `max()` slips through and resolves
to a bogus (out-of-range) index, reporting `error::not_found` rather than
`error::token_overflow`. Replaced with the pre-check
`result > (std::size_t(-1) - d) / 10`.

The regression test appends a digit to `max()` and expects `token_overflow`; it
fails before the change and passes after. The existing overflow test only covers
`max() + 1`, which happens to wrap below `result` and so was already caught.

### DIFF
--- a/include/boost/json/impl/pointer.ipp
+++ b/include/boost/json/impl/pointer.ipp
@@ -200,15 +200,19 @@ parse_number_token(
             return {};
         }
 
-        // guard against std::size_t overflow in result * 10 + d
-        if( result > (std::size_t(-1) - d) / 10 )
+        if( result > std::size_t(-1) / 10 )
         {
             BOOST_JSON_FAIL(ec, error::token_overflow);
             return {};
         }
+        result *= 10;
 
-        result = result * 10 + d;
-
+        if( result > std::size_t(-1) - d )
+        {
+            BOOST_JSON_FAIL(ec, error::token_overflow);
+            return {};
+        }
+        result += d;
     }
     return result;
 }
--- a/include/boost/json/impl/pointer.ipp
+++ b/include/boost/json/impl/pointer.ipp
@@ -200,14 +200,14 @@ parse_number_token(
             return {};
         }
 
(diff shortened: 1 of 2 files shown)


include/boost/json/impl/pointer.ipp
```diff
@@ -200,14 +200,14 @@ parse_number_token(
...
-        if( new_result < result )
+        // guard against std::size_t overflow in result * 10 + d
+        if( result > (std::size_t(-1) - d) / 10 )
         {
             BOOST_JSON_FAIL(ec, error::token_overflow);
             return {};
         }
 
-        result = new_result;
+        result = result * 10 + d;
```
grisumbras: How about something like this?

```
if( result > UINT64_MAX / 10 )
{
    BOOST_JSON_FAIL(ec, error::token_overflow);
    return {};
}
result *= 10;

if( result > UINT64_MAX - d )
{
    BOOST_JSON_FAIL(ec, error::token_overflow);
    return {};
}
result += d;
```

I've checked on Godbolt and produces less code.

Ramya-9353: Done, pushed the two-step guard. One adjustment: result is std::size_t, so I used std::size_t(-1) as the bound rather than UINT64_MAX, which would never fire on 32-bit targets. Pointer suite still passes, 185 total, including the max()+'9' regression.

grisumbras:
Thank you for catching this. Merged as e2dde54f6ad4bb8cd82730772106aba30edd76b0.

PR #1172, opened 2026-07-15
use array storage in insert(pos, first, last)

Ramya-9353:
Repro: back an array with a non-default resource and insert a forward-iterator range, e.g. `array a(&mr); a.insert(a.end(), v.begin(), v.end());` with `mr` a `monotonic_resource`. The inserted elements report a `storage()` different from `a.storage()`.
Cause: the forward-iterator `insert` overload constructs each element as `value(*first++)`, copying with the source value's storage rather than the array's `sp_`. Both range constructors, the input-iterator `insert` overload, and the count-fill overload already pass `sp_`; this path was the one that did not.
Fix: pass `sp_` so inserted elements adopt the array's storage.

Old behaviour also leaks: with a trivially-deallocatable resource the array destructor fast path skips element destructors, so any heap allocation an inserted element owns is never freed.

The regression test extends the existing random-iterator insert case to check element storage; it fails before the change and passes after.

### DIFF
--- a/include/boost/json/impl/array.hpp
+++ b/include/boost/json/impl/array.hpp
@@ -565,7 +565,7 @@ insert(
     revert_insert r(pos, n, *this);
     while(n--)
     {
-        ::new(r.p) value(*first++);
+        ::new(r.p) value(*first++, sp_);
         ++r.p;
     }
     return r.commit();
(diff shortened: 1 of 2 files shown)


grisumbras:
Thanks for catching this. This is quite a nasty bug.

PR #1174, opened 2026-07-17
buffer initializer_list before reallocation in array::insert

Ramya-9353:
Repro: `a.insert(a.begin(), {a[0], a[1]})` on an array with `size() == capacity()` reports an ASAN heap-use-after-free; the `initializer_list<value_ref>` overload builds `revert_insert` first, which on the growth path frees the old table before `write_array` dereferences the `value_ref`s that point into it.
Fix: materialise the init-list into a temporary array before constructing `revert_insert`, then relocate, matching the input-iterator insert path. The regression test inserts a self-referential init-list into a full array; it reports a heap-use-after-free under ASAN before the fix and passes after.

### DIFF
--- a/include/boost/json/impl/array.ipp
+++ b/include/boost/json/impl/array.ipp
@@ -516,10 +516,21 @@ insert(
         value_ref> init) ->
     iterator
 {
+    BOOST_ASSERT(
+        pos >= begin() && pos <= end());
+    if(init.size() == 0)
+        return data() + (pos - data());
+    // the value_refs in init may point into this
+    // array, whose storage revert_insert can
+    // relocate and free, so buffer them first
+    array temp(init, sp_);
     revert_insert r(
-        pos, init.size(), *this);
-    value_ref::write_array(
-        r.p, init, sp_);
+        pos, temp.size(), *this);
+    relocate(
+        r.p,
+        temp.data(),
+        temp.size());
+    temp.t_->size = 0;
     return r.commit();
 }
 
--- a/include/boost/json/impl/array.ipp
+++ b/include/boost/json/impl/array.ipp
@@ -516,10 +516,6 @@ insert(
         value_ref> init) ->
     iterator
 {
-    BOOST_ASSERT(
-        pos >= begin() && pos <= end());
-    if(init.size() == 0)
-        return data() + (pos - data());
     // the value_refs in init may point into this
     // array, whose storage revert_insert can
     // relocate and free, so buffer them first
--- a/test/array.cpp
+++ b/test/array.cpp
@@ -1035,18 +1035,35 @@ public:
         }
 
         // insert(const_iterator, init_list)
-        fail_loop([&](storage_ptr const& sp)
         {
-            array a({0, 3, 4}, sp);
-            auto it = a.insert(
-                a.begin() + 1, {1, str_});
-            BOOST_TEST(it == a.begin() + 1);
-            BOOST_TEST(a[0].as_int64() == 0);
-            BOOST_TEST(a[1].as_int64() == 1);
-            BOOST_TEST(a[2].as_string() == str_);
-            BOOST_TEST(a[3].as_int64() == 3);
-            BOOST_TEST(a[4].as_int64() == 4);
-        });
+            fail_loop([&](storage_ptr const& sp)
+            {
+                array a({0, 3, 4}, sp);
+                auto it = a.insert(
+                    a.begin() + 1, {1, str_});
+                BOOST_TEST(it == a.begin() + 1);
+                BOOST_TEST(a[0].as_int64() == 0);
+                BOOST_TEST(a[1].as_int64() == 1);
+                BOOST_TEST(a[2].as_string() == str_);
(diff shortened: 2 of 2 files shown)


Ramya-9353:
Rebased onto current develop to re-run CI. The only red stage on the previous drone build (Linux GCC 12 arm64) hung for ~44h and was killed by the runner; all other 65 stages passed, including the ASan/UBSan/TSan/Valgrind jobs. No code changes.

grisumbras:
I thought, I commented on this PR, but it appears that I have forgotten to do that. While I understand the problem, in this case the solution might be to just declare such use UB and hence motivate the caller to create the temporary themselves.

Did you catch the bug in an organic usage scenario, or have you encountered it simply by analysing corner cases?

Ramya-9353:
Corner-case analysis, not an organic report. After the fill-value aliasing fix in #1167 I went through the remaining array mutators looking for reads of user-supplied sources after reallocation, and this overload was the last one left.

On declaring it UB: that's defensible, but it would make this overload the odd one out. The count-fill and input-iterator overloads already tolerate aliasing, and the equivalent braced-list insert on std::vector is well defined because the list materialises copies before the call, so the pointer semantics of value_ref are what make it surprising here; nothing at the call site hints that a temporary is needed. The cost is one temporary array on this overload only, the same as the range insert already pays. That said, if you'd rather document it as a precondition instead, I can rework the PR that way.

grisumbras:
Almost the entire point of this overload is to avoid an allocation of a temporary. Otherwise you could just do `ja.insert(ja.end(), array{1, 2, 3})`. Creating a temporary in its implementation begs the question if the overload should exist at all.

Ramya-9353:
Fair enough, the temporary does undercut the reason the overload exists. Reworked as you suggested: the code change and test are gone, and the doc block now declares the aliasing UB with a precondition next to the existing `first`/`last` one on the range overload. Callers who do need the self-referential form can materialise the temporary themselves.

grisumbras:
I was persuaded that my cost-benenfit analysis was wrong. Can you please return the PR to fix UB instead of documenting it?

Ramya-9353:
Done. Dropped the doc-only commit and restored the code fix: the init-list is buffered into a temporary array before revert_insert runs, same shape as the input-iterator overload, with an early return for the empty list. The regression test (self-referential init-list insert into a full array) is back too; it reports a heap-use-after-free under ASan on develop and the full array suite passes clean with the fix, ASan/UBSan.

Ramya-9353:
Rebased onto current develop to pick up the GCC 16 -Wfree-nonheap-object fix for null_resource.cpp, which was the only thing failing on the drone build (three GCC 16 stages, all in that one test). No changes to the patch itself; array suite still passes under ASan/UBSan after the rebase.

test/array.cpp
```diff
@@ -1035,18 +1035,35 @@ class array_test
...
+                BOOST_TEST(a[0].as_int64() == 0);
+                BOOST_TEST(a[1].as_int64() == 1);
+                BOOST_TEST(a[2].as_string() == str_);
+                BOOST_TEST(a[3].as_int64() == 3);
+                BOOST_TEST(a[4].as_int64() == 4);
+            });
+
+            // elements of init alias *this and
+            // insertion reallocates
+            fail_loop([&](storage_ptr const& sp)
```
grisumbras: Just add the contents of this loop to the previous one.

include/boost/json/impl/array.ipp
```diff
@@ -516,10 +516,21 @@ insert(
         value_ref> init) ->
     iterator
 {
+    BOOST_ASSERT(
+        pos >= begin() && pos <= end());
+    if(init.size() == 0)
+        return data() + (pos - data());
```
grisumbras: This line is not tested.
https://1174.json.prtest2.cppalliance.org/diff-report/include/boost/json/impl/array.ipp.html#NL522

It begs the question, if this entire branch is necessary.

include/boost/json/impl/array.ipp
```diff
@@ -516,10 +516,21 @@ insert(
         value_ref> init) ->
     iterator
 {
+    BOOST_ASSERT(
```
grisumbras: This assert is not necessary, the iterator's validity is already asserted elsewhere.

grisumbras (dismissed):

include/boost/json/impl/array.ipp
Ramya-9353: Dropped. revert_insert's constructor already asserts pos is within [begin, end], so this was a duplicate.

include/boost/json/impl/array.ipp
Ramya-9353: Removed the branch. An empty list gives n_==0, which takes the fast path in revert_insert and early-returns without relocating, so commit() hands back the right iterator on its own. The special case was dead.

test/array.cpp
Ramya-9353: Done, folded the aliasing case into the existing fail_loop.

PR #1175, opened 2026-07-19
fix overlapping memcpy in string::assign

Ramya-9353:
## Summary

`string::assign(char const* s, size_type count)` uses
`char_traits::copy` (memcpy) unconditionally.  When `s` points into
the string's own buffer — e.g. via `s.assign(string_view(s).substr(3))` —
the source and destination regions overlap, which is undefined behaviour
per [mem.res]/[char.traits.require].

Under AddressSanitizer this fires `memcpy-param-overlap`; without
instrumentation it silently corrupts the result on some compilers/optimisation
levels.

## Fix

Detect aliasing with `detail::ptr_in_range`.  When the source is internal,
use `char_traits::move` (memmove) directly followed by `impl_.term(count)`.
The non-aliasing path remains unchanged.

The aliasing case guarantees `count <= size() <= capacity()`, so no
reallocation is needed and bypassing `impl_.assign` is safe.

## Reproducer

```cpp
boost::json::string s("abcdefghijklmnopqrstuvwxyz0123456789");
s.assign(boost::json::string_view(s).substr(3));
// ASAN: memcpy-param-overlap
// Without ASAN: silent data corruption (null terminator overwrites source)
```

## Labels

- component: `boost/json/string`
- type: `memcpy-param-overlap` / undefined behaviour
- severity: medium — data corruption on common self-referencing patterns
- affected: `string::assign(char const*, size_type)` and all callers
  (`operator=(string_view)`, `assign(string_view)`, construction from
  self-referencing `string_view`)

### DIFF
--- a/include/boost/json/impl/string.ipp
+++ b/include/boost/json/impl/string.ipp
@@ -176,9 +176,18 @@ assign(
     char const* s,
     size_type count)
 {
-    std::char_traits<char>::copy(
-        impl_.assign(count, sp_),
-        s, count);
+    auto const p = impl_.data();
+    if(detail::ptr_in_range(p, p + impl_.size(), s))
+    {
+        std::char_traits<char>::move(p, s, count);
+        impl_.term(count);
+    }
+    else
+    {
+        std::char_traits<char>::copy(
+            impl_.assign(count, sp_),
+            s, count);
+    }
     return *this;
 }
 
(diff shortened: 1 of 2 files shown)


Ramya-9353:
any update?

grisumbras:
Thank you for catching this. I slightly changed the tests and merged via c3a56f4c6edfe5ef86c0e905c5e6609ecefea218.

PR #1176, opened 2026-07-21
fix string corruption in string_impl::shrink_to_fit

Ramya-9353:
Repro: build a heap-allocated `string` (size past the SBO), shrink its size to at most the SBO capacity, then call `shrink_to_fit()`; the contents come back as unrelated bytes. Second case: shrink a heap string that stays above the SBO capacity, then read `c_str()`, and `strlen` exceeds `size()`.

Cause: in the heap-to-SBO branch `s_.k` is set to `short_string_` before `data()` is read, so once the kind has flipped `data()` returns the inline SBO buffer and the memcpy copies that buffer onto itself rather than the heap characters, leaving stale union bytes (the old table pointer among them). In the heap-to-smaller-heap branch only `size()` bytes are copied, so the terminator is dropped and `c_str()` reads into uninitialised storage.

Fix: read the source pointer before switching the kind, and copy `size() + 1` bytes on the reallocating branch so the terminator is carried across.

### DIFF
--- a/include/boost/json/detail/impl/string_impl.ipp
+++ b/include/boost/json/detail/impl/string_impl.ipp
@@ -435,9 +435,10 @@ shrink_to_fit(
     auto const t = p_.t;
     if(t->size <= sbo_chars_)
     {
+        char const* const p = data();
         s_.k = short_string_;
         std::memcpy(
-            s_.buf, data(), t->size);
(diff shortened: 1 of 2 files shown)


grisumbras:
Thank you for catching this serious issue. Merged via #1177.

PR #1178, opened 2026-07-30
reuse existing capacity in sbo_buffer::grow

Ramya-9353:
Repro: a 200-element `[1.25,...]` array (1001 bytes) fed to `stream_parser` a byte at a time with `number_precision::precise`, so every number straddles a `write_some` boundary, makes 813 global `operator new` calls totalling 56.1 GB with a largest single block of 1.14 GB; a 32 KB body in 4 KiB chunks peaks at a 2.2 MB block, and that peak keeps doubling as the body grows.
Cause: `grow` never checks whether the buffer already has room, so every `append` reallocates and each one floors the new capacity at `old_capacity * 2`, while `clear` keeps the block, so the doubling carries across numbers and across `reset`, and the inline buffer goes unused past the first append.
Fix: return early when the requested size fits the current capacity, matching what `string_impl::append` already does; the two runs above then total 19.8 kB and 240 kB, with largest blocks of 6 kB and 48 kB.

### DIFF
--- a/include/boost/json/detail/sbo_buffer.hpp
+++ b/include/boost/json/detail/sbo_buffer.hpp
@@ -151,6 +151,9 @@ public:
         }
 
         std::size_t const old_capacity = this->capacity();
+        if( size <= old_capacity - size_ )
+            return;
+
         std::size_t new_capacity = size_ + size;
 
         // growth factor 2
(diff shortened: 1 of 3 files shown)


grisumbras:
Thank you for catching this.
Slightly changed and merged via 5779b03b185f9623a59c3dd72a974db02943e92a.

PR #1179, opened 2026-08-03
fix deallocate argument in monotonic_resource

Ramya-9353:
Repro: parse anything through a `monotonic_resource` whose upstream is `test/checking_resource.hpp`. Every block fails `size == it->second.first`.

Cause: `do_allocate` asks the upstream for `sizeof(block) + next_size_` but records only `next_size_` in `block::size`, so `release()` returns each block 32 bytes short on LP64. The default resource ignores the size argument, so this only bites upstreams that bucket by size.

Fix: add the header size back before deallocating. The key-string alignment fix that was originally part of this PR moved to #1181.

### DIFF
--- a/include/boost/json/impl/monotonic_resource.ipp
+++ b/include/boost/json/impl/monotonic_resource.ipp
@@ -103,7 +103,7 @@ release() noexcept
     while(p != &buffer_)
     {
         auto next = p->next;
-        upstream_->deallocate(p, p->size);
+        upstream_->deallocate(p, sizeof(block) + p->size);
         p = next;
     }
     buffer_.p = reinterpret_cast<
--- a/test/monotonic_resource.cpp
+++ b/test/monotonic_resource.cpp
@@ -18,6 +18,7 @@
 #include <boost/core/max_align.hpp>
 #include <iostream>
 
+#include "checking_resource.hpp"
 #include "test_suite.hpp"
 
 namespace boost {
@@ -306,11 +307,23 @@ R"xx({
     }
 
     void
+    testAllocation()
+    {
+        // every block obtained from the upstream resource has to be
+        // returned to it with the size it was allocated with
+        BOOST_TEST_CHECKPOINT();
+        checking_resource res;
+        monotonic_resource mr(1, &res);
+        (void)mr.allocate(1, alignof(core::max_align_t));
+    }
+
+    void
     run()
     {
         testMembers();
         testStorage();
         testGeneral();
+        testAllocation();
     }
 };
 


include/boost/json/detail/string_impl.hpp
```diff
@@ -267,7 +267,8 @@ class string_impl
             // kind increases the cost of the destructor.
             // This function should be skipped when using
             // monotonic_resource.
-            sp->deallocate(k_.s, k_.n + 1);
+            sp->deallocate(k_.s, k_.n + 1,
+                alignof(char));
         }
```
grisumbras, line 271: While this is a correct fix, it is unrelated to the topic of this commit. Please move to a separate PR.

Btw, you don't have to add a line here, the extra argument perfectly fits into 80 columns, which is what I use as the threshold.

test/monotonic_resource.cpp
```diff
@@ -305,12 +307,42 @@ R"xx({
...
 
+    void
+    testAllocation()
+    {
+        // every block obtained from the upstream resource has to be
+        // returned to it with the size it was allocated with
+        {
+            checking_resource res;
+            monotonic_resource mr(1024, &res);
+            mr.allocate(1000, alignof(core::max_align_t));
```
grisumbras: https://drone.cpp.al/boostorg/json/2138/1/2

Silence these warnings with a cast to void (e.g. `(void)mr.allocate...`

test/monotonic_resource.cpp
```diff
@@ -305,12 +307,42 @@ R"xx({
...
+            mr.allocate(1000, alignof(core::max_align_t));
+            mr.allocate(4000, alignof(core::max_align_t));
+        }
+
+        {
```
grisumbras: This test does not appear to be necessary.

test/parse.cpp
```diff
@@ -202,13 +203,33 @@ class parse_test
         BOOST_TEST_THROWS_WITH_LOCATION( parse(ss) );
     }
 
+    void
+    testAllocation()
```
grisumbras: I presume, this should also be a part of a separate PR centered on requesting correct alignment in string deallocation.

grisumbras (dismissed):
Thank you for catching these serious issues.

include/boost/json/detail/string_impl.hpp
Ramya-9353: Done, moved it to #1181 and kept the call on one line there.

test/monotonic_resource.cpp
Ramya-9353: Done, cast to void.

test/monotonic_resource.cpp
Ramya-9353: Dropped it, the direct allocate calls already cover the block bookkeeping.

test/parse.cpp
Ramya-9353: Yes, moved to #1181 along with the alignment fix.

Ramya-9353:
Trimmed this down to just the monotonic_resource size fix, with the warnings silenced and the parse-through-monotonic test dropped. The key-string alignment fix and its test are now #1181.

test/monotonic_resource.cpp
```diff
@@ -305,12 +306,25 @@ R"xx({
...
+    void
+    testAllocation()
+    {
+        // every block obtained from the upstream resource has to be
+        // returned to it with the size it was allocated with
+        checking_resource res;
+        monotonic_resource mr(1024, &res);
+        (void)mr.allocate(1000, alignof(core::max_align_t));
+        (void)mr.allocate(1000, alignof(core::max_align_t));
+        (void)mr.allocate(4000, alignof(core::max_align_t));
```
grisumbras: I don't think you need the 3 allocates to manifest the problem. Locally I could do it with
```
monotonic_resource mr(1, &res);
(void)mr.allocate(1, alignof(core::max_align_t));
```
So, I think, this is how the test should be.
Also, add `BOOST_TECT_CHEKPOINT()` before it, so that the test runner outputs where the error originates.

include/boost/json/impl/monotonic_resource.ipp
```diff
@@ -103,7 +103,8 @@ release() noexcept
     while(p != &buffer_)
     {
         auto next = p->next;
-        upstream_->deallocate(p, p->size);
+        upstream_->deallocate(p,
+            sizeof(block) + p->size);
```
grisumbras: No need to make it 2 lines.

test/monotonic_resource.cpp
Ramya-9353: Done, reduced to the single allocate with mr(1, &res) and added BOOST_TEST_CHECKPOINT() above it. Verified it still fails against the old release(), and the runner now reports the checkpoint line.

include/boost/json/impl/monotonic_resource.ipp
Ramya-9353: Done, back on one line.

grisumbras:
Please, rebase on current develop and push again, so that CI relaunches. Also, squash into one commit.

Ramya-9353:
Done, rebased on current develop and squashed to a single commit. CI should be off again.

PR #1181, opened 2026-08-05
fix deallocate alignment in string_impl::destroy

Ramya-9353:
Split out from #1179 as requested.

Repro: parse a truncated document like `{"some key here":` with `test/checking_resource.hpp` as the resource. The abandoned key fails `alignment == it->second.second`.

Cause: `string_impl::destroy` frees a key string with the default alignment (max_align), but both `key_t` constructors allocate it with `alignof(char)`. The default resource ignores the argument, so this only bites upstreams that bucket by alignment.

Fix: pass `alignof(char)` to match the allocation. Both test cases fail before the fix and pass after.

### DIFF
--- a/test/parse.cpp
+++ b/test/parse.cpp
@@ -15,6 +15,7 @@
 
 #include <string>
 
+#include "checking_resource.hpp"
 #include "test.hpp"
 #include "test_suite.hpp"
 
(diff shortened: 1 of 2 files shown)


test/parse.cpp
```diff
@@ -202,13 +203,33 @@ class parse_test
...
+        // a key that is abandoned because the document is truncated still
+        // has to be returned to the resource with the size and alignment
+        // it was allocated with
+        {
+            checking_resource res;
+            system::error_code ec;
+            parse(R"({"some key here":)", ec, &res);
+        }
+
+        {
```
grisumbras: These two tests appear to test for the same exact thing. So, just remove one of them.

Ramya-9353: Done, dropped the second one. They both go through the same single-view key_t constructor anyway, so one case covers it.

PR #1182, opened 2026-08-06
fix required member check for described structs in parse_into

Ramya-9353:
Repro: `parse_into` a described struct whose members are all required, from `{"a": 1, "a": 2, "a": 3}`. It reports success, and `b` and `c` keep whatever the caller's object held. Same for `{"a": 2}` as the second element of `[{"a": 1, "b": 1, "c": "one"}, {"a": 2}]` parsed into `std::vector<X>`.
Cause: `activated_` counts values signalled rather than distinct members, and is never cleared, so a repeated key stands in for a missing one, and each object in a sequence starts from the previous object's count.
Fix: record which members have been set, and clear that record in `on_object_begin`, the way the sequence handler already clears its container in `on_array_begin`.

### DIFF
--- a/include/boost/json/detail/parse_into.hpp
+++ b/include/boost/json/detail/parse_into.hpp
@@ -18,6 +18,7 @@
 #include <boost/json/value.hpp>
 #include <boost/describe/enum_from_string.hpp>
 
+#include <array>
 #include <vector>
 
 /*
@@ -1153,6 +1154,7 @@ private:
     handler_tuple<converting_handler, InnerHandlers> handlers_;
     int inner_active_ = -1;
     std::size_t activated_ = 0;
+    std::array<bool, mp11::mp_size<Dt>::value> seen_ = {};
 
 public:
     converting_handler( converting_handler const& ) = delete;
@@ -1183,8 +1185,11 @@ public:
         bool required_member = mp11::mp_with_index<InnerCount>(
             inner_active_,
             is_required_checker{});
-        if( required_member )
+        if( required_member && !seen_[inner_active_] )
+        {
+            seen_[inner_active_] = true;
             ++activated_;
+        }
 
         key_ = {};
         inner_active_ = -1;
@@ -1214,7 +1219,11 @@ public:
     bool on_object_begin( system::error_code& ec )
     {
         if( inner_active_ < 0 )
+        {
+            activated_ = 0;
+            seen_ = {};
             return true;
+        }
 
         BOOST_JSON_INVOKE_INNER( on_object_begin(ec) );
     }
--- a/include/boost/json/detail/parse_into.hpp
+++ b/include/boost/json/detail/parse_into.hpp
@@ -1153,7 +1153,6 @@ private:
 
     handler_tuple<converting_handler, InnerHandlers> handlers_;
     int inner_active_ = -1;
-    std::size_t activated_ = 0;
     std::bitset<mp11::mp_size<Dt>::value> seen_;
 
 public:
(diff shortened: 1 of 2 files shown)


grisumbras:
I will have to think a bit about this one. I'm not sure we should check for this, because non-unique keys are explicitly discouraged by the RFC. On the other hand, I see the value this adds.

include/boost/json/detail/parse_into.hpp
```diff
@@ -1153,6 +1154,7 @@ class converting_handler<described_class_conversion_tag, V, P>
     handler_tuple<converting_handler, InnerHandlers> handlers_;
     int inner_active_ = -1;
     std::size_t activated_ = 0;
+    std::array<bool, mp11::mp_size<Dt>::value> seen_ = {};
```
grisumbras: This should be an `std::bitset` instead.

Ramya-9353: Done, switched to std::bitset (and the include from <array> to <bitset>). Suite still passes, 9666 assertions, 0 failures.

Ramya-9353:
On the RFC point: fair, though the patch does not actually reject duplicate keys as such. A document with repeated keys still parses so long as every required member is supplied somewhere; a repeat just no longer counts as a second distinct member. The other half of the bug also triggers with entirely unique keys: `[{"a": 1, "b": 1, "c": "one"}, {"a": 2}]` into `std::vector<X>` was accepted with the second element's `b` and `c` never written, because `activated_` carried over the first object's count. Same carryover reaches map values and nested struct members.

include/boost/json/detail/parse_into.hpp
```diff
@@ -1183,8 +1185,11 @@ class converting_handler<described_class_conversion_tag, V, P>
         bool required_member = mp11::mp_with_index<InnerCount>(
             inner_active_,
             is_required_checker{});
-        if( required_member )
+        if( required_member && !seen_[inner_active_] )
+        {
+            seen_[inner_active_] = true;
             ++activated_;
```
grisumbras: It appears that  with this change `seen_` completely replaces `activated_`. So, please remove the latter.

After you do that, the check can once again be just `if( required_member )`.

Ramya-9353: Done. activated_ is gone, on_object_end compares seen_.count() against the required-member count, and signal_value is back to a plain if( required_member ). Suite still passes, 9666 assertions, 0 failures.

include/boost/json/detail/parse_into.hpp
```diff
@@ -1234,7 +1229,7 @@ class converting_handler<described_class_conversion_tag, V, P>
         {
             using C = mp11::mp_count_if<Dt, is_optional_like>;
             constexpr int N = mp11::mp_size<Dt>::value - C::value;
-            if( activated_ < N )
+            if( seen_.count() < N )
```
grisumbras: Shouldn't this be `! seen_.all()`?

grisumbras:
Also, please rebase on current develop, so that CI stops failing.

include/boost/json/detail/parse_into.hpp
Ramya-9353: Done. count() < N was only there because optional members never set their bit, so all() would have failed for any struct with an optional member. on_object_begin now marks optional members as seen when it clears the record, so the check is just !seen_.all() and the N computation is gone. Suite passes, 9666 assertions, 0 failures, and a struct with an absent std::optional member still parses.

Ramya-9353:
Rebased onto current develop.

grisumbras:
After some thinking I reconsidered my request to check `seen_.all()` and preferred your earlier version. Now merged as e0e74f634194855b4daea49ed4016ef67aa38bff. Thanks again for catching this.

PR #1184, opened 2026-08-08
fix dropped storage in single-element value(initializer_list)

Ramya-9353:
Repro: value jv({{1, 2, 3}}, sp) reports jv.storage() as the default resource rather than sp; the two-or-more element and object list forms adopt sp correctly.
Cause: the size()==1 branch of value(initializer_list<value_ref>, storage_ptr) constructs *this on the default resource with scalar(), builds the element on sp, then swaps; swap keeps each operand's own storage, so sp is discarded.
Fix: construct *this directly from the element's value, which already holds sp.

### DIFF
--- a/include/boost/json/impl/value.ipp
+++ b/include/boost/json/impl/value.ipp
@@ -240,9 +240,8 @@ value(
     {
         if( init.size() == 1 )
         {
-            ::new(&sca_) scalar();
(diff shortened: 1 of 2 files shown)


grisumbras:
Thank you for catching this. Merged as ccdd27e484dfb0578e2d793a947bdab010a2bd5a.

PR #1185, opened 2026-08-11
account for completed utf-8 sequence in parse_string size limit

Ramya-9353:
Repro: feed a multi-byte UTF-8 string to `stream_parser` one byte at a time, so each character completes across a `write_some` boundary; `max_string_size`/`max_key_size` is not applied to it and `on_string_part`/`on_key_part` report a `total` that omits those bytes.
Cause: the `str8` resume path in `parse_string` (completing a UTF-8 sequence split across a chunk) emits `seq_.length()` bytes to the handler without bounds-checking them or adding them to `total`. `do_str1` and `parse_escaped` both account before their handler calls; `str8` was the gap.
Fix: on the `str8` path, check `seq_.length()` against the remaining budget and add it to `total` before emitting, matching `do_str1`.

Regression test in `test/limits.cpp` drives split multi-byte strings and keys through `stream_parser` a byte at a time; the size limits are skipped before the change and enforced after.

### DIFF
--- a/include/boost/json/basic_parser_impl.hpp
+++ b/include/boost/json/basic_parser_impl.hpp
@@ -1110,6 +1110,7 @@ do_str8:
     }
     {
         std::size_t const n = seq_.length();
+        bool r;
         if(is_key)
         {
             BOOST_ASSERT(total <= Handler::max_key_size);
@@ -1120,6 +1121,8 @@ do_str8:
                     = BOOST_CURRENT_LOCATION;
                 return fail(cs.begin(), error::key_too_large, &loc);
             }
+            total += n;
+            r = h_.on_key_part( {seq_.data(), n}, total, ec_ );
         }
         else
         {
@@ -1131,11 +1134,9 @@ do_str8:
                     = BOOST_CURRENT_LOCATION;
                 return fail(cs.begin(), error::string_too_large, &loc);
             }
+            total += n;
+            r = h_.on_string_part( {seq_.data(), n}, total, ec_ );
         }
-        total += n;
-        bool const r = is_key?
-            h_.on_key_part( {seq_.data(), seq_.length()}, total, ec_ ):
-            h_.on_string_part( {seq_.data(), seq_.length()}, total, ec_ );
         if(BOOST_JSON_UNLIKELY( !r ))
             return fail( cs.begin() );
     }
--- a/include/boost/json/basic_parser_impl.hpp
+++ b/include/boost/json/basic_parser_impl.hpp
@@ -1109,6 +1109,30 @@ do_str8:
         return fail(cs.begin(), error::syntax, &loc);
     }
     {
+        std::size_t const n = seq_.length();
+        if(is_key)
+        {
+            BOOST_ASSERT(total <= Handler::max_key_size);
+            if(BOOST_JSON_UNLIKELY(n >
+                Handler::max_key_size - total))
+            {
+                BOOST_STATIC_CONSTEXPR source_location loc
+                    = BOOST_CURRENT_LOCATION;
+                return fail(cs.begin(), error::key_too_large, &loc);
+            }
+        }
+        else
+        {
+            BOOST_ASSERT(total <= Handler::max_string_size);
+            if(BOOST_JSON_UNLIKELY(n >
+                Handler::max_string_size - total))
+            {
+                BOOST_STATIC_CONSTEXPR source_location loc
+                    = BOOST_CURRENT_LOCATION;
+                return fail(cs.begin(), error::string_too_large, &loc);
+            }
+        }
+        total += n;
         bool const r = is_key?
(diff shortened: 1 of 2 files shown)


test/limits.cpp
```diff
@@ -320,6 +320,63 @@ class limits_test
...
+                    {
+                        p.write_some(&c, 1, ec);
+                        if(ec)
+                            return;
+                    }
+                    p.finish(ec);
+                };
+
+            // string: one character past the limit
+            {
```
grisumbras: I think, the test is too complex for its own good. What it should do
1. Feed the parser with prefix (just `"\""` for a string value).
2. Feed the parser with `string::max_size()` ASCII characters (any one will do).
3. Feed the parser with a UTF-8 multibyte character.

At every step check that no error occurred. After step 3 check that the proper error has occurred.  No need for temporary buffers and helper functions.

test/limits.cpp
```diff
@@ -320,6 +320,63 @@ class limits_test
...
+                    body += "\xC3\xA9";
+                body += "\"";
+                system::error_code ec;
+                feed_split(body, ec);
+                BOOST_TEST(ec == error::string_too_large);
+                BOOST_TEST(ec.has_location());
+            }
+
+            // key: one character past the limit
+            {
```
grisumbras: Same as for the first test.

test/limits.cpp
```diff
@@ -320,6 +320,63 @@ class limits_test
...
+                    body += "\xC3\xA9";
+                body += "\":null}";
+                system::error_code ec;
+                feed_split(body, ec);
+                BOOST_TEST(ec == error::key_too_large);
+                BOOST_TEST(ec.has_location());
+            }
+
+            // a split multi-byte string within the limit still parses
+            {
```
grisumbras: I don't think this test is needed. I am pretty sure we test that UTF-8 characters are supported in principle.

test/limits.cpp
```diff
@@ -320,6 +320,63 @@ class limits_test
...
+                        if(ec)
+                            return;
+                    }
+                    p.finish(ec);
+                };
+
+            // string: one character past the limit
+            {
+                std::string body = "\"";
+                for(std::size_t i = 0;
```
grisumbras: Please, don't split lines where it's not necessary. That complicates reading the code.

include/boost/json/basic_parser_impl.hpp
```diff
@@ -1109,6 +1109,30 @@ parse_string(const char* p,
...
+            if(BOOST_JSON_UNLIKELY(n >
+                Handler::max_string_size - total))
+            {
+                BOOST_STATIC_CONSTEXPR source_location loc
+                    = BOOST_CURRENT_LOCATION;
+                return fail(cs.begin(), error::string_too_large, &loc);
+            }
+        }
+        total += n;
         bool const r = is_key?
```
grisumbras: It feels like this should be merged with the if-else block that you've added.

grisumbras (changes requested):

test/limits.cpp
Ramya-9353: Done, reworked to that shape: opening quote, then string::max_size() ASCII characters in one write, then the multibyte character. One deviation from your sketch: the final character has to arrive split across two writes (0xC3, then 0xA9). If both bytes land in the same write it's consumed inline by do_str1, whose existing check already fires, and the test passes even without the fix; the str8 resume path only runs for a sequence broken across a boundary.

test/limits.cpp
Ramya-9353: Done, same shape as the string test.

test/limits.cpp
Ramya-9353: Dropped.

test/limits.cpp
Ramya-9353: Fixed; the loops are gone with the rewrite.

include/boost/json/basic_parser_impl.hpp
Ramya-9353: Done, the size check, total update and handler call now live in one if-else and the ternary is gone.

grisumbras:
So, coverage checks fail on this PR here: https://1185.json.prtest2.cppalliance.org/diff-report/include/boost/json/basic_parser_impl.hpp.html#NL1125

This shows that we do not have a test for when a key has a UTF-8 multibyte character, the parser suspends within it and then resumes. Can you please add a test for that as a separate commit, so that the PR's CI is green?

Ramya-9353:
Added in a separate commit: same feed shape as the key overflow test, but with string::max_size() - 2 filler bytes so the split character completes on the str8 resume path exactly at the limit and the key parses cleanly. That runs the on_key_part call the diff report flagged as uncovered. Passes locally under ASan/UBSan.

grisumbras:
Merged as 5fad60c7e5968e5b25e30b1f270d0e1f2dee13e1 and c5273748ccbfbb29a044437d8ca0d95602b03d5a.

PR #1186, opened 2026-08-15
forward mapped value in object::insert_or_assign assign branch

Ramya-9353:
Repro: `insert_or_assign(key, std::move(v))` on an existing key, with `v` on the object's own resource, deep-copies `v` and leaves it intact instead of moving it. For a 100-element array the assign branch runs 101 allocations and the source array is still populated afterwards.
Cause: the assign branch casts the mapped value to `M` rather than `M&&`, so for an rvalue argument it selects `value(value const&)`, which allocates a copy on the source's storage before the copy into `sp_`. The insert branch one line above already forwards with `M&&`.
Fix: cast to `M&&` to match the insert branch. The stored value is unchanged; the redundant copy and the ignored move go away.

### DIFF
--- a/test/object.cpp
+++ b/test/object.cpp
@@ -1201,6 +1201,17 @@ public:
                     { "k3", 3 },
                     { "k4", 2 }}));
             }
+
+            // rvalue is moved, not copied, on the assign branch
+            {
+                object o = {
+                    { "k1", 1 },
+                    { "k2", { 1, 2, 3 } } };
+                value v = { 4, 5, 6 };
+                o.insert_or_assign("k2", std::move(v));
+                BOOST_TEST(o.at("k2") == (array{4, 5, 6}));
+                BOOST_TEST(v.as_array().empty());
+            }
         }
 
         // emplace(key, arg)
--- a/test/object.cpp
+++ b/test/object.cpp
@@ -1208,9 +1208,10 @@ public:
                     { "k1", 1 },
                     { "k2", { 1, 2, 3 } } };
                 value v = { 4, 5, 6 };
+                auto const d = v.as_array().data();
                 o.insert_or_assign("k2", std::move(v));
                 BOOST_TEST(o.at("k2") == (array{4, 5, 6}));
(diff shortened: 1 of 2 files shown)


grisumbras:
Ouch. Thanks for catching this.

test/object.cpp
```diff
@@ -1201,6 +1201,17 @@ class object_test
...
+
+            // rvalue is moved, not copied, on the assign branch
+            {
+                object o = {
+                    { "k1", 1 },
+                    { "k2", { 1, 2, 3 } } };
+                value v = { 4, 5, 6 };
+                o.insert_or_assign("k2", std::move(v));
+                BOOST_TEST(o.at("k2") == (array{4, 5, 6}));
+                BOOST_TEST(v.as_array().empty());
```
grisumbras: I think, a better check for whether `v` was actually moved is storing `v.as_array().data()` and comparing it to `o["k2"].as_array().data()`.

Ramya-9353: Agreed, comparing buffer identity pins down the move directly instead of relying on the moved-from state. Done: the test now stores v.as_array().data() before the call and checks it equals o.at("k2").as_array().data() afterwards. Verified it discriminates: with the old static_cast<M> the pointer differs (fresh copy), with M&& it's preserved.

test/object.cpp
```diff
@@ -1201,6 +1201,18 @@ class object_test
...
+
+            // rvalue is moved, not copied, on the assign branch
+            {
+                object o = {
+                    { "k1", 1 },
+                    { "k2", { 1, 2, 3 } } };
+                value v = { 4, 5, 6 };
+                auto const d = v.as_array().data();
+                o.insert_or_assign("k2", std::move(v));
+                BOOST_TEST(o.at("k2") == (array{4, 5, 6}));
```
grisumbras: I don't think this line is necessary at all.

Ramya-9353: Right, the pointer check subsumes it: same buffer implies same contents. Dropped.

PR #1188, opened 2026-08-18
unlink rolled-back keys from hash buckets on insert rollback

Ramya-9353:
Repro: bulk-insert a range or initializer_list into a hash-mode object that already has spare capacity, where a later element throws (e.g. an allocation failure) after an earlier one has been inserted; a subsequent lookup reads a freed key (ASAN heap-use-after-free in `find_in_object`).

Cause: on the no-reallocation path `revert_insert::destroy` rolls back the partially inserted elements and the size, but `insert_impl` has already linked each of them into a bucket chain, and those bucket heads are left pointing at the slots that are then destroyed.

Fix: unlink the rolled-back elements from their buckets in `revert_insert::destroy`, on the no-reallocation path only, before they are destroyed. Both bulk-insert overloads share this path. Regression test in `test/object.cpp` fails under ASAN before the change and passes after.

### DIFF
--- a/include/boost/json/impl/object.ipp
+++ b/include/boost/json/impl/object.ipp
@@ -200,9 +200,10 @@ destroy() noexcept
             obj_->remove( obj_->t_->bucket( last->key() ), *last );
         }
     }
-    obj_->destroy(
-        &(*obj_->t_)[size_],
-        obj_->end());
+    if(! obj_->sp_.is_not_shared_and_deallocate_is_trivial())
+        obj_->destroy(
+            &(*obj_->t_)[size_],
+            obj_->end());
 }
 
 //----------------------------------------------------------
--- a/include/boost/json/impl/object.ipp
+++ b/include/boost/json/impl/object.ipp
@@ -186,6 +186,20 @@ object::
 revert_insert::
 destroy() noexcept
 {
+    // when no reallocation happened, insert_impl linked each rolled-back
+    // element into a bucket of the live table; unlink them while their keys
+    // are still valid, otherwise a bucket head is left pointing at a slot
+    // that is about to be destroyed
+    if( !t_ && !obj_->t_->is_small() )
+    {
+        key_value_pair* const first = &(*obj_->t_)[size_];
+        key_value_pair* last = obj_->end();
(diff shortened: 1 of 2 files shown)


grisumbras:
I have managed to reliably reproduce this without needing an ASAN or adding elements in a loop.

It goes like this:

```c++
// static_resource is important because it doesn't deallocate before its destructor
unsigned char buf[1024];
static_resource mr(buf, sizeof(buf));

object jo(&mr);
jo.reserve(20); // reserve more than "small object"

jo["1"] = 1; // add one element so that we remain not empty

std::array<throws_on_convert, 3> input; // array of things convertible to key_value_pair
input[0].k = "2";
input[0].should_throw = false;

input[1].k = "3";
input[1].should_throw = false;

input[2].k = "4";
input[2].should_throw = true; // third element throws on conversion

BOOST_TEST_THROWS(
  jo.insert( input.begin(), input.end() ),
  std::invalid_argument); // check that we indeed throw

// stale buckets from insert have info for jo["2"]
// which points to jo.begin() + 2 , which is currently jo.end() + 1
BOOST_TEST( jo.find("3") != jo.end() );
```

Please change the test in the PR to something like this. This also reveals another bug, which will require a separate commit.

Ramya-9353:
Done. I've replaced the test with your repro (gave throws_on_convert a key member so the rolled-back keys are distinct), asserting find("2")/find("3") == end() after the throw. Without the unlink it fails exactly as you describe, with find("3") returning an iterator past the end.

The other bug it surfaces is revert_insert::destroy calling object::destroy unconditionally, which trips the BOOST_ASSERT for non-shared trivial-deallocate storage like static_resource. Guarded it the same way clear() does, in a separate commit.

Ramya-9353:
Rebased onto develop to pick up 82398f9. The three red Drone stages (GCC 16, GCC static linking, Valgrind GCC 16) were all the -Wfree-nonheap-object false positive in test/null_resource.cpp rather than anything in this diff; the sanitiser stages were green. Diff is unchanged.

grisumbras:
Once again, thank you for catching this. This was a very serious bug.

PR #1191, opened 2026-08-21
account for alignment padding in monotonic_resource::do_allocate

Ramya-9353:
Repro: `monotonic_resource mr; mr.allocate(1024, 64);` returns a null pointer in release builds and trips `BOOST_ASSERT(p)` in debug (UBSan: "applying non-zero offset 1024 to null pointer" at monotonic_resource.ipp:145). The same happens through `pmr::polymorphic_allocator<T>` for any `T` with `alignof(T) > alignof(max_align_t)` once a pmr vector grows to a power-of-two byte size.
Cause: `do_allocate` sizes a fresh block from `n` alone, but the block payload is only `alignof(block)` aligned, so `std::align` on the new block has no room for the padding an over-aligned request needs whenever `n + padding > next_size_`.
Fix: include the worst-case padding (`align - alignof(block)`) when choosing the block size, and reject a padded size that would overflow with `bad_alloc`, as `static_resource` already does. The added test allocates 1024 bytes at alignments from `2 * max_align` to 4096 from fresh resources; it fails 7 assertions before the change and passes after.

### DIFF
--- a/include/boost/json/impl/monotonic_resource.ipp
+++ b/include/boost/json/impl/monotonic_resource.ipp
@@ -14,8 +14,10 @@
 #include <boost/json/monotonic_resource.hpp>
 #include <boost/json/detail/except.hpp>
 #include <boost/core/max_align.hpp>
+#include <boost/throw_exception.hpp>
 
 #include <memory>
+#include <new>
 
 namespace boost {
 namespace json {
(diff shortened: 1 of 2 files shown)


Ramya-9353:
gentle ping

grisumbras:
This PR has exposed that our memory resources do not support overaligned types. I'll have to look into how this is treated in e.g. Boost.Container first, before I decide whether to merge this.

PR #1192, opened 2026-08-24
fix array index overflow in set_at_pointer

Ramya-9353:
Repro: call `set_at_pointer` on a non-empty array with an index token equal to `std::size_t(-1)`, e.g. `"/18446744073709551615"` on a 64-bit build, and a `max_created_elements` large enough to pass the element-count guard.

Cause: the guard bounds `n` (the number of elements to create) but not `index + 1`. `arr.size() + n + 1` is `index + 1`, which wraps to `0`, so `resize` takes the shrink branch (skipping the `array_too_large` check) and the lambda returns `arr.data() + index`, a pointer one element before the buffer. The following `*result = ref.make_value(...)` then writes through it (ASAN: heap-buffer-overflow in `value::operator=` via pointer.ipp:490; UBSan: unsigned pointer overflow at pointer.ipp:462).

Fix: reject `index >= array::max_size()` before the resize and fail with `error::array_too_large`, next to the existing `max_created_elements` check. Regression test drives index `std::size_t(-1)` on a non-empty array and checks the array is untouched and `error::array_too_large` is returned.

### DIFF
--- a/include/boost/json/impl/pointer.ipp
+++ b/include/boost/json/impl/pointer.ipp
@@ -455,11 +455,13 @@ value::set_at_pointer(
                 if( n >= opts.max_created_elements )
                     return nullptr;
 
-                // arr.size() + n + 1 equals index + 1; reject the value that
-                // would wrap so the resize count and the returned pointer stay
-                // in bounds
-                if( index == std::size_t(-1) )
+                // the resulting size, index + 1, would either exceed the
+                // array's size limit or wrap to 0 when index == size_t(-1)
+                if( index >= array::max_size() )
+                {
+                    BOOST_JSON_FAIL( ec, error::array_too_large );
                     return nullptr;
+                }
 
                 arr.resize( arr.size() + n + 1 );
             }
--- a/include/boost/json/impl/pointer.ipp
+++ b/include/boost/json/impl/pointer.ipp
@@ -455,6 +455,12 @@ value::set_at_pointer(
                 if( n >= opts.max_created_elements )
                     return nullptr;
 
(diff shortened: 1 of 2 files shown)


include/boost/json/impl/pointer.ipp
```diff
@@ -455,6 +455,12 @@ value::set_at_pointer(
                 if( n >= opts.max_created_elements )
                     return nullptr;
 
+                // arr.size() + n + 1 equals index + 1; reject the value that
+                // would wrap so the resize count and the returned pointer stay
+                // in bounds
+                if( index == std::size_t(-1) )
+                    return nullptr;
```
grisumbras, line 463: This will result in the function returning a "not found" error. I don't think this is fitting. I'd rather have it return something related to the idea that we cannot create an array this big.

Which also makes me think that it's better to compare against the maximum size of the array (`index >= array::max_size()`). And set `ec` to `error::array_too_large`.

Ramya-9353: Makes sense, array_too_large describes the failure better. Changed the check to `index >= array::max_size()` with `BOOST_JSON_FAIL(ec, error::array_too_large)` and updated the test to expect that error. Pointer tests still pass under ASan/UBSan.

Ramya-9353:
Rebased onto develop, no change to the diff. The three red Drone stages (GCC 16, GCC static linking, Valgrind GCC 16) were failing on -Wfree-nonheap-object in test/null_resource.cpp, which 82398f9 fixed on develop after my last push.

grisumbras:
Thank you for catching and fixing this.

PR #1197, opened 2026-09-13
Fix build

bkuhls:
The build errors were found with gcc-15.3.0/uClibc-ng/x86_64 by the Gitlab pipelines of the buildroot project:
https://gitlab.com/bkuhls/buildroot/-/jobs/16471960148


grisumbras:
The linked log shows Charconv build failing, not Json.

bkuhls:
> The linked log shows Charconv build failing, not Json.

correct, that was the build job that found multiple errors, the first one in charconv, after fixing that other build errors in json occured. The commit messages in this PR show the real build errors.

grisumbras:
Thanks for catching and fixing this.

PR #1200, opened 2026-09-20
clear accumulated name in parse_into enum handler

Ramya-9353:
Repro: `parse_into` a `std::vector<E>` of a described enum from `["x", "y"]`; it fails with `error::unknown_name`. With `enum class role { user, admin, useradmin }`, `["user", "admin"]` parses as `{ user, useradmin }`. Escapes are not needed when reading from a stream: `parse_into(v, is)` feeds the parser in `BOOST_JSON_STACK_BUFFER_SIZE` chunks, so a plain `"user","admin"` whose first string straddles a chunk boundary gives the same `{ user, useradmin }`.
Cause: the described enum handler buffers a string delivered in parts in `name_`, and whenever `name_` is non-empty it looks up `name_` plus the final part. `name_` is never emptied after a lookup, and the same handler object serves every element of a sequence, every mapped value of a map and the same member of every struct in a sequence, so each later enum string is looked up with the earlier name prepended.
Fix: clear `name_` after a successful lookup, before signalling the parent, since the variant handler destroys the alternative's handler inside `signal_value`. The added `std::vector<E>` round trip fails in the byte-at-a-time `parser_for` loop of `testParseIntoValue` before the change (14 failed assertions) and passes after.


grisumbras:
Thank you for catching and fixing this.

